PR #5, opened 2021-01-28
chore: added bee-js package from npmjs

vojtechsimetka:


nugaon:
github test workflow alternation is needed, set `BEE_API_URL` that the tests will use.

vojtechsimetka:
Can't we unify this between this project and `bee-js`? So either use `BEE_API_URL` or `BEE_URL`

agazso (approved):

nugaon:
> Can't we unify this between this project and `bee-js`? So either use `BEE_API_URL` or `BEE_URL`

I suggest to use BEE_API_URL in `bee-js` as well, because debug api functions could come into the picture

PR #6, opened 2021-01-31
feat: identity

nugaon:
# Description
This PR adds `identity` commands to the CLI.
It handles wallets of the user, which will be mostly used for sign chunks.

Config feature also has been added, in order to store wallets on the user's computer.
The configuration file is placed in a hidden folder named `swarm-cli`.
In case of Unix-based systems this config path will be: `$HOME/.swarm-cli`
On Windows systems: `$HOME\AppData\swarm-cli`

The configuration file is saved with 600 file permission.

There are two types of wallets currently in this feat PR: `simple` and `V3 keystore`.
- The first only generates a private key, that it will store in the config fire in cleartext. These can be generated quickly.
- The latter is Ethereum compatible wallet, which fits to the `V3` standard. Its creation is slower and also requires a password.

If wallet import functionality is needed directly, (because it is possible now, by altering the just mentioned config file) please describe your idea below.

## TODO
- [x] Write tests
- [ ] Add import possibility 
- [x] Review README
- [x] Add `bee-js` dependency

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@
 
 > Command line interface tool for manage Bee node and utilize its functionalities
 
-The goal of this project to handle all Swarm related operations through CLI in the future.
+The goal of this project is to handle most of the Swarm operations through CLI at some point in the future.
 For currently supported operations, see [Commands](##Commands) section.
 
 ## Commands
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ The assigment priority how option gets its value in question is the following:
 With specific system environment variables you can alter the behaviour of the CLI
 
 * `BEE_API_URL` - API URL of Bee client
-* `SWARM_CLI_CONFIG_FOLDER` - full path of configuration folder
+* `SWARM_CLI_CONFIG_FOLDER` - full path to a configuration folder
 
 # Install
 
--- a/README.md
+++ b/README.md
@@ -43,8 +43,8 @@ It is possible to set value of particular parameters in different ways.
 The assignment priority of how option gets its value in question is the following:
 
 1. passed CLI option value (with e.g. `--option-example-1`)
-2. env variable for that option in form of either `OPTION_EXAMPLE_1` or `SWARM_CLI_OPTION_EXAMPLE_1` (if it's available)
-3. CLI configuration value of that option (if it's available)
+2. env variable for that option in form of either `OPTION_EXAMPLE_1` or `SWARM_CLI_OPTION_EXAMPLE_1` (if it is available)
+3. CLI configuration value of that option (if it is available)
 4. option's default fallback value (or it is required to define by #1)
 
 ## System environment
--- a/README.md
+++ b/README.md
@@ -9,39 +9,87 @@
 ![](https://img.shields.io/badge/Node.js-%3E%3D10.0.0-orange.svg?style=flat-square)
 ![](https://img.shields.io/badge/runs%20in-browser%20%7C%20node%20%7C%20webworker%20%7C%20electron-orange)
 
-> Cli tool for manage Bee node and utilize its functionalities
-
 **Warning: This project is in alpha state. There might (and most probably will) be changes in the future to its API and working. Also, no guarantees can be made about its stability, efficiency, and security at this stage.**
 
-## Install
+# Description
 
-### npm
+> Command line interface tool for manage Bee node and utilize its functionalities
+
+The goal of this project to handle all Swarm related operations through CLI in the future.
+For currently supported operations, see [Commands](##Commands) section.
+
+## Commands
+
+`swarm-cli` can perform commands:
+- upload # upload files (even directories) to Swarm network by passing the file's path
+- identity # manage keys (which can be compatible with Ethereum V3 keystore standard) that you can use mostly for signing chunks
+
+for more info execute `swarm-cli --help` after installation.
+
+## Config
+
+The configuration file is placed in a hidden folder named swarm-cli.
+In case of Unix-based systems this config path will be: $HOME/.swarm-cli
+On Windows systems: $HOME\AppData\swarm-cli
+
+The configuration file is saved with 600 file permission.
+
+On first run, this configuration will be generated with default values, that you are able to change on your demand under the before mentioned path.
+
+## Assignment priority
+
+It is possible to set value of particular parameters in different ways.
+
+The assigment priority how option gets its value in question is the following:
+
+1. passed CLI option value (with e.g. `--option-example-1`)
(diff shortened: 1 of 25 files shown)


src/service/identity/index.ts
```diff
@@ -0,0 +1,51 @@
...
+    case IdentityType.simple:
+      return 'only keypair'
+    case IdentityType.v3:
+      return 'v3 keystore'
+    default:
+      throw new Error(`IdentityType '${identityType}' is not known.`)
+  }
+}
+
+export function isSimleWallet(wallet: IdentityWallet, identityType: IdentityType): wallet is SimpleWallet {
```
agazso, line 16: Typo: `isSimleWallet` -> `isSimpleWallet`

src/command/root-command.ts
```diff
@@ -12,17 +14,31 @@ export class RootCommand {
...
-    this.beeApiUrl = this.beeApiUrl === defaultBeeApiUrl ? process.env.BEE_API_URL || this.beeApiUrl : this.beeApiUrl
+    this.commandConfig = new CommandConfig(this.appName, this.configFolder)
+    this.beeApiUrl = this.optionPassed(beeApiUrl)
+      ? this.beeApiUrl
+      : this.commandConfig.config.beeApiUrl || process.env.BEE_API_URL || this.beeApiUrl
     this.bee = new Bee(this.beeApiUrl)
   }
+
+  protected optionPassed(option: IOption): boolean {
+    return Boolean(process.argv.indexOf(`--${option.key}`) > -1)
```
agazso, line 42: Consider using `includes` instead of `indexOf`.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes

src/command/root-command.ts
```diff
@@ -12,17 +14,31 @@ export class RootCommand {
...
+  public commandConfig!: CommandConfig
+
   /**
    * Init Root command fields
    *
    * if BEE_API_URL environment variable has been set the CLI will use that connection string
    */
   protected init(): void {
-    this.beeApiUrl = this.beeApiUrl === defaultBeeApiUrl ? process.env.BEE_API_URL || this.beeApiUrl : this.beeApiUrl
+    this.commandConfig = new CommandConfig(this.appName, this.configFolder)
```
agazso, line 34: I find the logic here a bit difficult to follow. If I understand correctly the `CommandConfig` constructor is either
- creates a config file (if there were none) with the default values (but potentially overridden with options)  and returns that 
- or it reads the config file with the values in it

If there is no config it creates a config (for example with the `beeApiUrl`) with the value of the provided option. Therefore in the subsequent invocations it will always use the value of the first invocation unless you pass in an option. But it seems it won't be saved, so the only way to change that is by editing the config file, right?

src/command/root-command.ts
```diff
@@ -12,17 +14,31 @@ export class RootCommand {
...
+    this.beeApiUrl = this.optionPassed(beeApiUrl)
```
agazso, line 35: What's the point of assigning the value to itself (`this.beeApiUrl = this.beeApiUrl`)?
Both branches fall back to this. Maybe this can be simplified.

agazso (changes requested):

src/command/root-command.ts
nugaon: Regardless of the first passed `beeApiUrl` option, it puts the `beeApiUrl`'s default fallback value to the default config file.

After your comment I was thinking about it, and maybe would be better to rearrange the assigment priority in the following way:
1. passed cli option value
2. env variable
3. config
4. option's default value

I would keep generate the default config with the default values of cli options that the cli anyway would fallback to. It is useful to indicate what kind of variables could be set via config, and how.

src/command/root-command.ts
nugaon: Thanks for noticing, I change this conditional statement to a good old `if` one.

Also rearranged the priority in that way that I mentioned before

agazso (approved):

vojtechsimetka (approved):

README.md
```diff
@@ -9,39 +9,87 @@
...
+2. env variable for that option in form of either `OPTION_EXAMPLE_1` or `SWARM_CLI_OPTION_EXAMPLE_1` (if it's available)
+3. CLI configuration value of that option (if it's available)
+4. option's default fallback value (or it is required to define by #1)
+
+## System environment
+
+With specific system environment variables you can alter the behaviour of the CLI
+
+* `BEE_API_URL` - API URL of Bee client
+* `SWARM_CLI_CONFIG_FOLDER` - full path of configuration folder
```
vojtechsimetka, line 55:
```suggestion
* `SWARM_CLI_CONFIG_FOLDER` - full path to a configuration folder
```

README.md
```diff
@@ -9,39 +9,87 @@
...
-> Cli tool for manage Bee node and utilize its functionalities
-
 **Warning: This project is in alpha state. There might (and most probably will) be changes in the future to its API and working. Also, no guarantees can be made about its stability, efficiency, and security at this stage.**
 
-## Install
+# Description
+
+> Command line interface tool for manage Bee node and utilize its functionalities
+
+The goal of this project to handle all Swarm related operations through CLI in the future.
```
vojtechsimetka, line 18:
```suggestion
The goal of this project is to handle most of the Swarm operations through CLI at some point in the future.
```

README.md
```diff
@@ -9,39 +9,87 @@
...
+
+The configuration file is saved with 600 file permission.
+
+On first run, this configuration will be generated with default values, that you are able to change on your demand under the before mentioned path.
+
+## Assignment priority
+
+It is possible to set value of particular parameters in different ways.
+
+The assigment priority how option gets its value in question is the following:
```
vojtechsimetka, line 43:
```suggestion
The assignment priority of how option gets its value in question is the following:
```

README.md
```diff
@@ -9,39 +9,87 @@
...
+
+1. passed CLI option value (with e.g. `--option-example-1`)
+2. env variable for that option in form of either `OPTION_EXAMPLE_1` or `SWARM_CLI_OPTION_EXAMPLE_1` (if it's available)
+3. CLI configuration value of that option (if it's available)
```
vojtechsimetka, line 47:
```suggestion
2. env variable for that option in form of either `OPTION_EXAMPLE_1` or `SWARM_CLI_OPTION_EXAMPLE_1` (if it is available)
3. CLI configuration value of that option (if it is available)
```
In general it is discouraged to shorten words in official-ish documents

vojtechsimetka (approved):

PR #12, opened 2021-02-11
feat: feeds hack

agazso:


agazso:
This is implemented in #32 in a proper way

Issue #14, opened 2021-02-12
Error color (dark red) is hard to read on dark background


agazso:
For example in a typical terminal.

It would be better to use a brighter shade.

Issue #13, opened 2021-02-12
Output is very verbose


agazso:
It would be nice to have either quiet mode or -v verbosity switch

PR #15, opened 2021-02-15
refactor: console log

nugaon:
`console.log` becomes more structured within this PR.
* `console.log` now only accessible through `root-command` parent class by default 
* some message types have been defined (e.g. `error`, `info`, `dim`, `divider`...)
* auto message formats for message types (italic, bold, etc.) 
* 3 levels of verbosity for restrict output messages (Quiet, Normal, Verbose) with optimized console message calling

Resolves: #13, resolves: #14


src/command/upload.ts
```diff
@@ -129,11 +136,11 @@ export class Upload extends RootCommand implements LeafCommand {
     syncedBar.stop()
 
     if (!synced) {
-      console.warn(red('Data syncing timeout.'))
+      this.console.error('Data syncing timeout.')
 
       return false
     } else {
-      console.log(dim('Data has been synced on Swarm network'))
+      this.console.log(dim('Data has been synced on Swarm network'))
```
vojtechsimetka, line 143:
```suggestion
      this.console.dim('Data has been synced on Swarm network')
```

src/command/upload.ts
```diff
@@ -5,8 +5,9 @@ import * as Path from 'path'
 import { sleep } from '../utils'
 import { Tag } from '@ethersphere/bee-js/dist/types'
 import { SingleBar, Presets } from 'cli-progress'
-import { bold, green, dim, red } from 'kleur'
+import { bold, green, dim } from 'kleur'
```
vojtechsimetka, line 8: These should probably not be used here, right?
```suggestion
```

src/command/root-command/command-log.ts
```diff
@@ -0,0 +1,98 @@
...
+        break
+      default:
+        // quite
+        this.error = error
+        this.log = emptyFunction
+        this.info = emptyFunction
+        this.dim = emptyFunction
+        this.divider = emptyFunction
+    }
+  }
```
vojtechsimetka, line 64: IMHO there should be `--silent` option as well which really just logs in the important stuff like hash on upload or really nothing on download.

vojtechsimetka: Or maybe just provide `--log-level` attribute with following options: `error, log, info, dim` or numeric value `1-4`

src/command/root-command/command-log.ts
```diff
@@ -0,0 +1,98 @@
...
+        break
+      case VerbosityLevel.Normal:
+        this.error = error
+        this.log = log
+        this.info = info
+        this.divider = divider
+        this.dim = emptyFunction
+        break
+      default:
+        // quite
```
vojtechsimetka, line 57:
```suggestion
        // quiet
```
or maybe even better `silence`

agazso:
I tried out with the different verbosity levels:

Screenshot 2021-02-15 at 15 27 49

My suggestions are the following:
- In level 0 only print only the bzz link (e.g. `http://localhost:1633/bzz/<hash>`>
- In level 1 omit the dark grey text (`Data has been synced...`), this could be the default
- Only print the dark grey text in level 2 (verbose mode) but maybe change the color to something more readable.

src/command/upload.ts
nugaon: It uses "bold" and "green" still, the commit suggestion is wrong

agazso (approved):

src/command/root-command/command-log.ts
nugaon: silent is identical to "Quiet". If someone wants totally silent output, the output can be piped to `/dev/null`. it is nice to print error message regardless of the given verbosity level on error, especially if the cli exits with error code anyway.
The last that you mentioned is too rigid to implement. Other message types can come into the picture.

nugaon:
> In level 0 only print only the bzz link (e.g. http://localhost:1633/bzz/<hash>>

I think the hash-only could be more useful as it does not contain arbitrary information and also there would be other use-cases when the gateway link does not matter.

> In level 1 omit the dark grey text (Data has been synced...), this could be the default

I think it is not relevant information, the point of "Normal" verbosity mode is with minimal formatting (and with deterministic row count) I get the result of the operation, otherwise some error happened.

> Only print the dark grey text in level 2 (verbose mode) but maybe change the color to something more readable.

The point of the `dim` messages to be hidden in the context, these are only for debugging information and tracking of the progress.
`dim` messages are only visible in verbosity level 2. it was a typeo that I left in the code, thereby it showed also in verbosity level 1, @vojtechsimetka  already gave change suggestion for that to fix.

vojtechsimetka (approved):

PR #21, opened 2021-02-23
feat: identity import and export

Cafe137:


src/command/identity/export.ts
```diff
@@ -0,0 +1,77 @@
...
+import { printNoIdentitiesError, promptUserForIdentity } from './common'
+
+export class Export extends RootCommand implements LeafCommand {
+  // CLI FIELDS
+
+  public readonly name = 'export'
+
+  public readonly description = 'Export identity'
+
+  @Argument({ key: 'identity-name', describe: 'Name of the identity to be exported' })
```
agazso, line 14: If this argument is required (as opposed to optional) you need to pass in a `required: true` property. Then it will be written in the help as required and also the framework will complain if the user does not provide it.

src/command/identity/export.ts
```diff
@@ -0,0 +1,77 @@
...
+  public identityName!: string
+
+  public async run(): Promise<void> {
+    this.initCommand()
+    this.checkForNoIdentities()
+    const options = this.getV3Identities()
```
agazso, line 20: I think it would be better to call this `identities` instead of `options`

src/command/identity/import.ts
```diff
@@ -0,0 +1,113 @@
...
+import { RootCommand } from '../root-command'
+
+export class Import extends RootCommand implements LeafCommand {
+  // CLI FIELDS
+
+  public readonly name = 'import'
+
+  public readonly description = 'Import V3 wallet as a new identity'
+
+  @Argument({ key: 'path', describe: 'Path to the V3 wallet file' })
```
agazso, line 15: See my previous comment regarding the `required` properties.

src/command/identity/export.ts
```diff
@@ -0,0 +1,77 @@
...
+    this.checkForEmptyIdentities(options)
+    await this.ensureIdentityNameIsProvided(options)
+    this.checkForExistingIdentity()
+    const identity: Identity = this.commandConfig.config.identities[this.identityName]
```
agazso, line 24: If you use the `getWalletFromIdentity` function here it returns a `Wallet` which has a `toV3String` function to create a V3 JSON. Then it would be possible to export the `privateKey` based identities as well.

Maybe it's worth considering or for a next PR.

nugaon: I don't think it's a good idea, because those are disposable entities without password only should work in a Swarm session ideally.
Also as I know for V3 format a password is neccessary in order to encrypt private key, thereby you have to pass it and the output won't be the same as the source.

Cafe137: I see some scenarios where a disposable keypair ends up being used more than it was originally supposed to, and then it would be beneficial to convert it to a secure V3 wallet.

We can use something like `Wallet.fromPrivateKey(privateKey).toV3String(password)` with a new password the user has to confirm.

Not sure if this is the scope of the first iteration, but I support developing this in the near future.

src/command/identity/import.ts
Cafe137: Agreed, added the `required` property.

src/command/identity/export.ts
Cafe137: Agreed, renaming done.

src/command/identity/export.ts
Cafe137: This is purposely not required - it works the same way as the `identity remove` command. We show an option picker with V3 identities only where the end user can select an identity in a user friendly way. Do you agree with this approach?

src/command/identity/export.ts
agazso: > We can use something like `Wallet.fromPrivateKey(privateKey).toV3String(password)` with a new password the user has to confirm.

I think that's a good direction.

> Not sure if this is the scope of the first iteration, but I support developing this in the near future.

I agree, let's keep it simple for now and add this as a feature later.

src/command/identity/import.ts
```diff
@@ -0,0 +1,104 @@
...
+      exit(1)
+    }
+  }
+
+  private async ensurePasswordIsProvided(): Promise<void> {
+    if (!this.password) {
+      this.console.log('You have not defined the password with the "--password" option.')
+      this.password = await this.console.askForPassword('Please provide the password for this V3 Wallet')
+    }
+  }
```
nugaon, line 57: I think this function will be needed outside of this class as well when we use identities for example for `feed` uploads.

_See comment: root-command-inheritance_

src/command/identity/export.ts
```diff
@@ -0,0 +1,77 @@
...
+  /** Init additional properties of class, that are not handled by the CLI framework */
+  private initCommand(): void {
+    super.init()
+  }
+
+  private getV3Identities(): string[] {
+    return Object.entries(this.commandConfig.config.identities)
+      .filter(identity => identity[1].identityType === IdentityType.v3)
+      .map(identity => identity[0])
+  }
```
nugaon, line 38: Can you give me more context about these indexing on identity like this please?

src/command/identity/import.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  private exists(path: string): boolean {
+    try {
+      const stat = statSync(path)
+
+      return stat.isFile()
+    } catch {
+      return false
+    }
+  }
```
nugaon, line 103: please, put this to [src/utils](src/utils) with name `fileExists`

nugaon (commented):
Please add spinning loading output as well where is needed.

src/command/identity/export.ts
Cafe137: `Object.entries()` returns an array of `[key, value]` pairs. It is the love child of `Object.keys()` and `Object.values()`.

I need both informations:

`[1]` / `value` / _identity object_ - to check `identityType` if it is a V3 wallet
`[0]` / `key` / _name of the identity_ - to collect the possible choices for the end user

nugaon (approved):
LGTM!

agazso (approved):

PR #32, opened 2021-03-10
feat: feed upload, update and print commands

Cafe137:


src/command/upload.ts
```diff
@@ -119,15 +156,19 @@ export class Upload extends RootCommand implements LeafCommand {
...
-      const updateState = tag.synced
 
+      if (updateState !== tag.synced) {
+        i = 0
+        updateState = tag.synced
+      }
       syncedBar.update(updateState)
 
-      if (tag.processed === updateState) {
+      if (tag.total === tag.synced) {
```
nugaon:
```suggestion
      if (tag.total === updateState) {
```
though see my prev comment about `tag.total`

test/command/feed.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+  let consoleMessages: string[] = []
+
+  beforeAll(() => {
+    global.console.log = jest.fn(message => {
+      consoleMessages.push(message)
+    })
+    jest.spyOn(global.console, 'warn')
+    //set config environment variable
+    process.env.SWARM_CLI_CONFIG_FOLDER = configFolderPath
+    process.env.SWARM_CLI_CONFIG_FILE = configFileName
```
nugaon, line 19: does it not affect the other tests which runs parallely and also would use these environments?

nugaon (commented):
Regarding to `src/command/upload.ts` please consider using [`Aggregation` decorator of the new version of `furious-commander`](https://github.com/nugaon/furious-commander#aggregation)

src/command/upload.ts
```diff
@@ -31,6 +33,8 @@ export class Upload extends RootCommand implements LeafCommand {
   @Option({ key: 'tag-polling-trials', describe: 'After the given trials the tag polling will stop', default: 15 })
   public tagPollingTrials!: number
 
+  public uploadAsFileList = false
```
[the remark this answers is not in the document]
Cafe137: It is only intended to be used internally, set from another command for example.

It is a workaround for uploading single files as a single-element browser-like File arrays.

I would not spend too much time fleshing this out, since this will be removed ASAP.

src/command/upload.ts
```diff
@@ -119,15 +156,19 @@ export class Upload extends RootCommand implements LeafCommand {
     const pollingTime = this.tagPollingTime
     const pollingTrials = this.tagPollingTrials
     let synced = false
+    let updateState = 0
     const syncedBar = new SingleBar({}, Presets.rect)
-    syncedBar.start(tag.processed, 0)
+    syncedBar.start(tag.total, 0)
```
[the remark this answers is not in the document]
Cafe137: It works locally with a single node, and inside GitHub actions.

test/command/feed.spec.ts
Cafe137: That is a great question, and it looks like it should, however, the tests are stable and there are no race conditions with the configurations.

nugaon: it could be a solution, if you would just pass the `--config-folder` option with a brand-new created empty testfolder.

nugaon:
I got after compilation and calling `swarm-cli`:
Screenshot from 2021-03-16 19-40-11
node version: v12.18.3

src/command/feed/feed-command.ts
```diff
@@ -0,0 +1,98 @@
...
+    const writer = this.bee.makeFeedWriter('sequence', topic, wallet.getPrivateKey())
+    const { reference } = await writer.upload(chunkReference)
+    const manifest = await this.bee.createFeedManifest('sequence', topic, wallet.getAddressString())
+
+    if (this.verbosity === VerbosityLevel.Verbose) {
+      this.console.log(bold(`Chunk Reference -> ${green(chunkReference)}`))
+      this.console.log(bold(`Chunk Reference URL -> ${green(`${this.beeApiUrl}/files/${chunkReference}`)}`))
+      this.console.log(bold(`Feed Reference -> ${green(reference)}`))
+      this.console.log(bold(`Feed Manifest -> ${green(manifest)}`))
+    }
```
nugaon, line 42: This is against the principals of the `this.console.log` usage. you shouldn't check for any vebosity in the code usually, if you don't want to put it into `dim` message feel free to create one neutral one for `Verbose`  in the `command-log.ts`.
Anyway IMO these messages could appear in `Normal` verbosity as well.

src/command/feed/feed-command.ts
```diff
@@ -0,0 +1,98 @@
...
+    this.console.log(bold(`Feed Manifest URL -> ${green(`${this.beeApiUrl}/bzz/${manifest}/`)}`))
+
+    if (this.verbosity === VerbosityLevel.Quiet) {
+      this.console.important(manifest)
+    }
```
nugaon, line 47: Maybe the similar way as I mentioned in my earlier comment, you can define a neutral `quiet` function for the console messages in `command-log.ts`

src/command/feed/upload.ts
```diff
@@ -0,0 +1,54 @@
...
+    this.console.dim('Successfully uploaded to feed.')
+  }
+
+  private async runUpload(): Promise<string> {
+    const upload = new FileUpload()
+    const stats = await stat(this.path)
+
+    if (stats.isFile()) {
+      upload.uploadAsFileList = true
+      upload.indexDocument = this.path
```
nugaon, line 36: the path of the file is not surely a relative link originated from its parent folder it also could be a full path as well, thereby I don't think it can be good.

nugaon (changes requested):
Within the subsequent PR, we should reconsider how to handle properly connections between LeafCommand definitions.

src/command/feed/feed-command.ts
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  protected getTopic(): string | Topic {
+    if (!this.hashTopic) {
+      this.enforceValidHexTopic()
+    }
+
+    return this.hashTopic ? this.bee.makeFeedTopic(this.topic) : this.topic
```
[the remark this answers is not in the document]
Cafe137: Cool!

src/command/feed/upload.ts
Cafe137: This was handled with a `basename` on the other side, but for clarity I moved it here now.

nugaon (approved):
LGTM!

PR #34, opened 2021-03-16
docs: add usage to README

### LINKED ISSUE #20: Documentation in README
Add documentation of basic features to the README

agazso:
This is based on the branch `feat/feed` to make the review easier.

Closes #20

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,10 +36,6 @@ The general usage is to provide a command, optionally a subcommand, then argumen
 
 `swarm-cli command [subcommand] <arguments> [options]`
 
-For example for uploading a file or folder to Swarm and updating the feed:
-
-`swarm-cli feed upload --identity main --topic a --hash-topic --path dist/index.js --password "$PASSWORD"`
-
 Running a command with the `--help` option prints out the usage of a command.
 
 ## Commands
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -18,13 +18,60 @@
...
+Let's say we want to upload a directory of our website to Swarm and update a feed to point to the newest version. For updating a feed we would need to sign it with an Ethereum key, so first we need to create one with the `identity create` command:
+
+```
+swarm-cli identity create
+```
 
-for more info execute `swarm-cli --help` after installation.
+This command will ask for a password. After that a new identity is created (named `main`). Now we can use this identity to sign updates. It's also possible to import and export Ethereum JSON V3 format identities that works with other apps (e.g. wallets).
+
+For uploading to a feed we can use the `feed upload` command. It expects an `identity` to be provided along with the `password` belonging to it and the `path` of the directory (or file) we want to upload.
```
nugaon, line 66: Directories are files
```suggestion
For uploading to a feed we can use the `feed upload` command. It expects an `identity` to be provided along with the `password` that belongs to it and the `path` of the file (even of a directory) we want to upload.
```

README.md
```diff
@@ -18,13 +18,60 @@
...
+
+```
+swarm-cli feed upload --identity main --password my-secret-password --path dist
+```
+
+In this example we are uploading the content of the `dist` folder. If the uploading was successful the last printed line will contain a `Feed Manifest URL`. This URL can be opened in the browser. If the uploaded folder contains an `index.html` file then it will be automatically added as a default file to be displayed when visiting the URL.
+
+We can also put this URL into an ENS record or reverse proxy configuration, because it will stay the same when the feed is updated. The uploaded content can be found on the link in the line starting with `URL`. This will change every time the content is modified.
```
nugaon, line 74: I'm not sure the URL itself should be placed to ENS, as the bee-client host address varies according to the CLI setup (which is `http://localhost:1633` by default)

nugaon (commented):
I'm not a native English speaker, but I tried my best.

README.md
agazso: I'm not sure I understand what you mean, can you please clarify?

Please note that the example is about uploading a website which will be most likely a directory. Also I plan to change directory to folder everywhere to be more consistent with the user expectations too.

README.md
agazso: This is a good point, I will update it.

nugaon (approved):

PR #57, opened 2021-03-25
feat: aliases, verbose and quiet mode, feed print with public address

Cafe137:


src/command/feed/feed-command.ts
```diff
@@ -8,21 +8,22 @@ import { Identity } from '../../service/identity/types'
...
     defaultDescription: '32 zero bytes',
   })
   public topic!: string
 
-  @Option({ key: 'password', describe: 'Password for the wallet' })
+  @Option({ key: 'password', alias: 'P', describe: 'Password for the wallet' })
   public password!: string
 
-  @Option({ key: 'hash-topic', type: 'boolean', describe: 'Hash the topic to 32 bytes', default: false })
+  @Option({ key: 'hash-topic', alias: 'h', type: 'boolean', describe: 'Hash the topic to 32 bytes', default: false })
```
agazso, line 26: I think this is already an alias for `--help`

agazso:
Regarding the `feed print`, I find it strange that it requires the password when I provide the identity with a name.

```
% node dist/index.js feed print -i main                                         
There is no password passed for V3 wallet initialization
% node dist/index.js feed print -i 0x3a68612583c65a0388418c37eeee6df1027a7042 -q
fdf21a5556e968c4079ba6b06b16a6ec545c2ce189db867c0ad5f6a0d4004de6
```

But if I list the identities it shows the address of an identity provided with the name without password:

```
% node dist/index.js identity list     
List of your identities
============================================================================================================================================================================================================
Identity name 	 main
Identity type 	 v3 keystore
Address 	 0x3a68612583c65a0388418c37eeee6df1027a7042
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

It shows that the address is stored non-encrypted, so it should be available for the print command as well without password.

src/command/identity/import.ts
```diff
@@ -18,10 +18,10 @@ export class Import extends RootCommand implements LeafCommand {
   @Argument({ key: 'path', required: true, describe: 'Path to the V3 wallet file' })
   public path!: string
 
-  @Option({ key: 'identityName', describe: 'Name of the identity to be saved as' })
+  @Option({ key: 'identityName', alias: 'i', describe: 'Name of the identity to be saved as' })
```
nugaon, line 21: in `export.ts` it is defines with dashes. let's keep that convention in key string names instead of this one

src/command/feed/print.ts
```diff
@@ -26,4 +25,15 @@ export class Print extends FeedCommand implements LeafCommand {
...
+  private async getAddressString(): Promise<string> {
+    if (this.commandConfig.config.identities[this.identity]) {
+      await this.checkIdentity()
+      const wallet = await this.getWallet()
+
+      return wallet.getAddressString()
+    }
+
+    return this.identity
+  }
```
nugaon, line 38: This initializes the wallet which is not needed and causes error in case of V3 wallets without password. within the configuration you can extract this info without any wallet init.

nugaon (commented):
Please add command aliases as well like upload -> up; download -> down, etc.
you can do that defining `aliases` in the command class, e.g.
```ts
public readonly aliases = ['up']
```

vojtechsimetka:
Please rebase on master for the tests to work

nugaon (approved):
Please consider my head comment message in https://github.com/ethersphere/swarm-cli/pull/57#pullrequestreview-621182014

Anyway we discussed everything else in the pair programming session, so LGTM!

agazso:
> Please consider my head comment message in [#57 (review)](https://github.com/ethersphere/swarm-cli/pull/57#pullrequestreview-621182014)
> 
> Anyway we discussed everything else in the pair programming session, so LGTM!

I tested it and works well. I opened a separate [issue about the aliases](https://github.com/ethersphere/swarm-cli/issues/66). Let's merge this.

PR #68, opened 2021-03-31
feat: add aliases for common commands

Cafe137:


agazso (approved):

agazso (commented):
Please add to the description that this closes #68

src/command/feed/upload.ts
```diff
@@ -5,6 +5,8 @@ import { FeedCommand } from './feed-command'
 export class Upload extends FeedCommand implements LeafCommand {
   public readonly name = 'upload'
 
+  public readonly aliases = ['u', 'up']
```
nugaon, line 8: Maybe I would remove these command aliases since `feed` command has **up**load and **up**date in the same time.

src/command/identity/list.ts
```diff
@@ -1,13 +1,15 @@
...
 import { bold } from 'kleur'
 import { getPrintableIdentityType, getSimpleWallet, isSimpleWallet, isV3Wallet } from '../../service/identity'
+import { RootCommand } from '../root-command'
 
 export class List extends RootCommand implements LeafCommand {
   // CLI FIELDS
 
   public readonly name = 'list'
 
+  public readonly aliases = ['l', 'ls']
```
nugaon, line 11: I would just leave `ls`

src/command/upload.ts
```diff
@@ -16,6 +16,8 @@ export class Upload extends RootCommand implements LeafCommand {
 
   public readonly name = 'upload'
 
+  public readonly aliases = ['u', 'up']
```
nugaon, line 19: I would just leave `up` as old swarm-cli had before

src/command/feed/upload.ts
agazso: That's a good point, let's just remove the aliases here

nugaon (approved):

Issue #64, opened 2021-03-29
Skip feed update if the uploaded content has not changed


agazso:
There could be a feature (e.g. an extra option) to the `feed upload` command to skip the feed update if the hash of the uploaded content is the same as in the last feed update.

Issue #36, opened 2021-03-17
swarm-cli does not work with Bee running in gateway node


significance:
this means that it is necessary to run one's own node.

it would be great if swarm-cli could work with a remote gateway, eg. gateway.ethswarm.org

agazso:
It seems that the reason is that swarm-cli is using tags functionality to track the progress of the upload and that is not available on gateways.

One idea to do work around this limitation is to only use tags on localhost and not on other domains unless the user specifies.

significance:
yes, i think it could be worth using this to open the discussion on enabling tags in gateway mode by otherwise mitigating potential attack vectors. it's a very important feature I think.

agazso:
Relevant issue: https://github.com/ethersphere/bee/issues/1369

Cafe137:
It is now possible to upload to gateway using the `--skip-sync` flag

Issue #61, opened 2021-03-28
swarm-cli tries to sync even if Bee is running in standalone mode


significance:
this doesn't make sense! 🤔 

```
Waiting for file chunks to be synced on Swarm network...
 ■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■ 99% | ETA: INFs | 2014/202074fee8cc9dbac82a954f907d15a454337f8713286af090166ab8a715ec07c02b
Data syncing timeout.
```

agazso:
How about we add a `--no-sync` flag to the `upload` and `feed upload` commands? It would also serve as a workaround for this and #36.

significance:
yeah i think it's worthwhile. would be good to continue the conversation about allowing tags at the gateway too.

PR #71, opened 2021-04-12
feat: add --skip-sync flag to upload and feed upload

Cafe137:
Added a new `--skip-sync` flag to `upload` and `feed upload`.

It disables tags and waiting for network synchronisation.

Using this option makes it possible to upload to the public gateway.

### DIFF
--- a/src/command/upload.ts
+++ b/src/command/upload.ts
@@ -60,8 +60,12 @@ export class Upload extends RootCommand implements LeafCommand {
 
   public async run(): Promise<void> {
     this.initCommand()
-    let tag = await this.bee.createTag()
     let url: string
+    let tag: Tag | undefined
+
+    if (!this.skipSync) {
+      tag = await this.bee.createTag()
+    }
 
     if (!FS.existsSync(this.path)) {
       this.console.error(`Given filepath '${this.path}' doesn't exist`)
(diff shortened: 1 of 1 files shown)


agazso:
On the gateway nodes the `tag` related features are disabled so it would be good if the flag would also disable the calls to `createTag` and `retrieveTag`.

agazso (approved):

agazso:
Closes #64

agazso:
Closes #36 and #61

src/command/upload.ts
```diff
@@ -82,10 +94,17 @@ export class Upload extends RootCommand implements LeafCommand {
...
+    if (this.skipSync) {
+      this.console.info('Skipping synchronization')
+    } else if (tag) {
+      tag = await this.bee.retrieveTag(tag.uid)
+      const synced = await this.waitForFileSynced(tag)
+
+      if (!synced) return //error message printed before
+    } else {
+      this.console.error('Tag was not created')
+    }
```
nugaon, line 107: Could it happen theoritically? I mean `tag = await this.bee.createTag()` would throw an error if this happens, doesn't it?

nugaon (approved):

src/command/upload.ts
Cafe137: You are right.

I originally added it for better readability since it is not very obvious that when `this.skipSync` is turned off, then `tag` always exists, but it does not add much value.

PR #73, opened 2021-04-20
feat: add pinning commands

Cafe137:
Adds 3 commands under the group `pinning`:

- `pin`
- `unpin`
- `list`

These commands can be used to pin chunks in retrospect.

There is a known issue at the time of adding this - manifests with index-document cannot be pinned or unpinned. In this case a helpful message is printed.

### DIFF
--- a/test/command/feed.spec.ts
+++ b/test/command/feed.spec.ts
@@ -1,4 +1,4 @@
-import { existsSync, unlinkSync } from 'fs'
+import { existsSync, readFileSync, unlinkSync } from 'fs'
 import { cli } from 'furious-commander'
 import { join } from 'path'
 import { optionParameters, rootCommandClasses } from '../../src/config'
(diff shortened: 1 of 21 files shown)


Cafe137:
Known issue: when `--index-document` is specified, chunk traversal fails and API call dies with `pin bzz: cannot pin`.

test/command/feed.spec.ts
```diff
@@ -73,4 +73,37 @@ describe('Test Feed command', () => {
...
+
+  it('should print feed using address only', async () => {
+    // create identity
+    await cli({
+      rootCommandClasses,
+      optionParameters,
+      testArguments: ['identity', 'create', '--identity-name', 'test2', '--password', 'test'],
+    })
+    const config = JSON.parse(readFileSync('test/testconfig/feed.config.json').toString())
+    const address = config?.identities?.test2?.wallet?.address
```
nugaon, line 85: please extract the created address from console.log or from the command's instance (if it stores)

nugaon (commented):
Please provide `unpin` tests as well, with checking whether it successfully removed the pinning with `pinning list`

nugaon (approved):
LGTM!

Issue #77, opened 2021-04-23
Bee Debug API URL value set


nugaon:
You can only pass the Bee Debug API URL by CLI option parameter, but would be nice to define it by system variable and config parameter as well (as in case of Bee API URL you can do that).

PR #78, opened 2021-04-23
feat: bee debug api url config and env

nugaon:
Closes: #77


Cafe137 (approved):
Good catch.

There was an issue with the tests, since the `optionPassed` in RootCommand does not work with CLI options.

Workaround was specifying those using `process.env` in the tests, that way they are picked up since that is global.

PR #79, opened 2021-04-26
refactor: update cli engine

Cafe137:
This PR changes the underlying parser from yargs to a custom one, built for the needs of swarm-cli.

The motive is to reduce argv handling / parsing and CLI related code in swarm-cli, and focus on application logic instead.

In addition, the compiler flag `"useDefineForClassFields": true` has been added, so that the project can be used via `ts-node` which is crucial for proper debugging.


src/command/root-command/index.ts
```diff
@@ -30,57 +23,32 @@ export class RootCommand {
...
-
-    if (this.optionPassed(quiet)) {
-      this.verbosity = VerbosityLevel.Quiet
-    }
-
-    if (this.optionPassed(verbose)) {
-      this.verbosity = VerbosityLevel.Verbose
+    this.commandConfig = new CommandConfig(this.appName, this.console, this.configFolder)
+    const sourcemap = Utils.getSourcemap()
+    if (sourcemap[beeApiUrl.key] === 'default' && this.commandConfig.config.beeApiUrl) {
```
agazso, line 35: We have a linter rule to enforce blank lines before `if` statements. 

Now this is subjective but I could also imagine having more newlines inserted in general to separate and group lines of code according to their functionality.

Cafe137: I don't always turn ESLint off temporarily, but when I do, I surely forget to turn it back on before pushing code 🙂

src/printer.ts
```diff
@@ -0,0 +1,15 @@
+import { Printer } from 'furious-commander/dist/printer'
+import { bold } from 'kleur'
+import { CommandLog, VerbosityLevel } from './command/root-command/command-log'
+
+export function createPrinter(): Printer {
+  const commandLog = new CommandLog(VerbosityLevel.Normal)
```
nugaon, line 6: Would be nice to figure out smth in a subsequent PR to share the same command log instance that the root-command has.
maybe with handling  the passed verbosity flag separately somehow here or with other solution.

nugaon (approved):
LGTM!

PR #81, opened 2021-04-29
fix: handle not found and bad request pinning errors

Cafe137:
Checking for `Internal Server Error` during `pinning unpin` was a mistake originally. It either returns `Not Found` or `Bad Request`. This PR handles these cases with a nicer error message and adds a test case for that.

In `pinning pin`, `Internal Server Error` is correct, but the `Not Found` case also exists there which was not handled. This PR also adds that with a test.

### DIFF
--- a/src/command/pinning/pin.ts
+++ b/src/command/pinning/pin.ts
@@ -21,7 +21,7 @@ export class Pin extends RootCommand implements LeafCommand {
       this.console.error('Could not pin ' + this.address)
 
       if (error.message === 'Not Found') {
-        this.console.error('No root chunks found with that address.')
+        this.console.error('No root chunk found with that address.')
       } else if (error.message === 'Internal Server Error') {
         this.console.error(
           'Pinning root chunks which were uploaded with --index-document header are currently not supported.',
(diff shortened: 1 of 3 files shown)


src/command/pinning/pin.ts
```diff
@@ -18,8 +18,11 @@ export class Pin extends RootCommand implements LeafCommand {
       await this.bee.pinCollection(this.address)
       this.console.log('Pinned successfully')
     } catch (error) {
-      if (error.message === 'Internal Server Error') {
-        this.console.error('Could not pin ' + this.address)
+      this.console.error('Could not pin ' + this.address)
+
+      if (error.message === 'Not Found') {
+        this.console.error('No root chunks found with that address.')
```
nugaon, line 24:
```suggestion
        this.console.error('No root chunk found with that address.')
```

nugaon (approved):

src/command/pinning/pin.ts
Cafe137: Thanks, I agree it reads better that way.

PR #98, opened 2021-05-07
feat!: add postage stamp commands

Cafe137:
Adds Postage Stamp support with 3 new commands: `stamp buy`, `stamp list` and `stamp show`. Some existing commands are also affected by this: `upload`, `feed upload`, `feed update` and `feed print`.


src/command/stamp/show.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export class Show extends StampCommand implements LeafCommand {
+  public readonly name = 'show'
+
+  public readonly aliases = []
+
+  public readonly description = 'Show a specific postage stamp'
+
+  @Argument({ key: 'batch-id', description: 'Batch ID of the postage stamp', required: true })
+  public batchId!: string
```
Cafe137, line 12: Other commands simply call this `stamp`, naming should be more consistent.

src/command/feed/print.ts
```diff
@@ -27,7 +27,7 @@ export class Print extends FeedCommand implements LeafCommand {
     const addressString = this.address || (await this.getAddressString())
     const reader = this.bee.makeFeedReader('sequence', topic, addressString)
     const { reference, feedIndex, feedIndexNext } = await reader.download()
-    const manifest = await this.bee.createFeedManifest('sequence', topic, addressString)
+    const manifest = await this.bee.createFeedManifest(this.stamp, 'sequence', topic, addressString)
```
Cafe137, line 30: Read operation should not use stamp

agazso: I agree that printing should not require a stamp, however currently we cannot calculate the feed manifest address on the client side so we rely on calling the `createFeedManifest` which uploads the feed manifest.

Until we implement the feed manifest calculation on the client side, I see the following options:
1. We don't print the manifest address. This is not great :)
2. We always require the stamp for printing the feed. This may be surprising to the users but it will work and print the manifest address.
3. If a stamp is provided as an optional argument we print the feed manifest address, otherwise not. This can be confusing to explain, but the default behavior will be the way it is supposed to be used in the future.

What do you think?

Cafe137: I think this is not a blocker, and we should keep it in its current state. Make it require a stamp so it can print the manifest. That is probably the most important part of the `feed print` command.

Then, as part of the next steps, let's implement this the right way with a client side calculation.

src/command/stamp/show.ts
agazso: This came up with the `bee-js` too. The root cause is that `batch` and `stamp` is both used to indicate the same thing in Bee and we sticked to `postageBatch` in the JS API.

However I find `stamp` a bit more specific and understandable as a user facing concept, so I would use that (or maybe `stamp-id` as argument). This may create a bit confusion in the code, so it would be good to add a comment about it.

Cafe137: Pushed a followup for this just now, without seeing your comment.

I ended up using:

```
key: 'stamp',
description: 'Batch ID of the postage stamp to use'
```

Maybe `ID of the postage batch to use` is a better description based on your feedback? But then `stamp` stands out a bit, not being mentioned in the description at all. Wording is hard. 🙂

agazso: I would go with `ID of the postage stamp to use`

src/command/stamp/buy.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  public readonly description = 'Buy postage stamp'
+
+  @Option({ key: 'depth', description: 'Depth of the postage stamp', type: 'number', required: true, minimum: 17 })
+  public depth!: number
+
+  @Option({ key: 'amount', description: 'Amount of the postage stamp', type: 'bigint', required: true, minimum: 1 })
+  public amount!: bigint
+
+  @Option({ key: 'label', description: 'Label of the postage stamp' })
```
nugaon, line 17: The descriptions could be clearer about the terms like
amount -> `epresents the value per chunk, which depreciates over time with new mined blocks.`

src/command/stamp/buy.ts
```diff
@@ -0,0 +1,35 @@
...
+  public label!: string
+
+  // CLASS FIELDS
+
+  public postageBatchId!: string
+
+  public async run(): Promise<void> {
+    super.init()
+
+    this.console.info('Buying postage stamp. This may take a while.')
```
nugaon, line 27: please put a spinner here

src/command/stamp/show.ts
```diff
@@ -0,0 +1,23 @@
+import { Argument, LeafCommand } from 'furious-commander'
+import { StampCommand } from './stamp-command'
+
+export class Show extends StampCommand implements LeafCommand {
+  public readonly name = 'show'
+
+  public readonly aliases = []
```
nugaon, line 7: you can delete this line

src/command/upload.ts
```diff
@@ -23,6 +23,13 @@ export class Upload extends RootCommand implements LeafCommand {
   @Argument({ key: 'path', description: 'Path to the file or folder', required: true })
   public path!: string
 
+  @Option({
+    key: 'stamp',
+    description: 'ID of the postage stamp to use',
+    required: true,
+  })
+  public stamp!: string
```
nugaon, line 31: see my comment on `feed-command`

nugaon:
we should unify later the postage stamp usage in our tests, but I liked your solution : )

nugaon (approved):

PR #99, opened 2021-05-10
refactor: extract printing functions to avoid instantiation

Cafe137:
Continuation of https://github.com/ethersphere/swarm-cli/pull/79#discussion_r623905757

This PR extracts the different `console.log` printing methods so they can be used without instantiation.

They are used to construct `CommandLog` as well as `Printer` for furious-commander.

### DIFF
--- a/src/command/root-command/command-log.ts
+++ b/src/command/root-command/command-log.ts
@@ -1,4 +1,3 @@
-/* eslint-disable no-console */
 import { prompt } from 'inquirer'
 import { exit } from 'process'
 import { Printer } from './printer'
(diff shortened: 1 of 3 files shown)


src/command/root-command/command-log.ts
```diff
@@ -31,45 +30,34 @@ export class CommandLog {
...
-        this.log = log
-        this.info = info
-        this.divider = divider
-        this.dim = emptyFunction
+        this.error = Printer.error
+        this.quiet = Printer.emptyFunction
+        this.verbose = Printer.emptyFunction
+        this.log = Printer.log
+        this.info = Printer.info
+        this.divider = Printer.divider
```
agazso, line 49: In the `Verbose` and `default` blocks the `dim` comes before the `divider` but here it's the other way.

nugaon (approved):
LGTM!

PR #100, opened 2021-05-10
refactor: simplify reading config via sourcemap

Cafe137:
Continuation of https://github.com/ethersphere/swarm-cli/pull/79#discussion_r623898323

Instead of having duplicated `if` blocks for each option that may be specified via the swarm-cli config file, we just have to maintain an array: `CONFIG_OPTIONS`.

### DIFF
--- a/src/command/root-command/command-config.ts
+++ b/src/command/root-command/command-config.ts
@@ -7,6 +7,13 @@ import { Identity } from '../../service/identity/types'
 import { ConfigOption } from '../../utils/types/config-option'
 import { CommandLog } from './command-log'
 
+/**
+ * Options listed here will be read from the config file
+ * when they are not set explicitly or via `process.env`.
+ *
+ * `optionKey` is the kebab-case variant of the argument used in the parser, a.k.a. the key,
+ * `propertyKey` is the camelCase variant used in TypeScript command classes, a.k.a. the property.
+ */
 export const CONFIG_OPTIONS: ConfigOption[] = [
   { optionKey: 'bee-api-url', propertyKey: 'beeApiUrl' },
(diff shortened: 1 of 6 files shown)


Cafe137:
Actually, we still don't have tests for this behaviour, so let me write a few first.

test/command/config.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+  const configFilePath = join(configFolderPath, configFileName)
+
+  beforeAll(() => {
+    global.console.log = jest.fn(message => {
+      consoleMessages.push(message)
+    })
+    jest.spyOn(global.console, 'warn')
+    //set config environment variable
+    process.env.SWARM_CLI_CONFIG_FOLDER = configFolderPath
+    process.env.SWARM_CLI_CONFIG_FILE = configFileName
```
nugaon, line 20: does it not conflict with the other test runs?

src/command/root-command/command-config.ts
```diff
@@ -4,8 +4,14 @@ import { join } from 'path'
 import { exit } from 'process'
 import { beeApiUrl, beeDebugApiUrl } from '../../config'
 import { Identity } from '../../service/identity/types'
+import { ConfigOption } from '../../utils/types/config-option'
 import { CommandLog } from './command-log'
 
+export const CONFIG_OPTIONS: ConfigOption[] = [
```
nugaon, line 10: Please write jsdoc about its role and how to handle

nugaon (approved):
LGTM!

test/command/config.spec.ts
Cafe137: Test suites are supposed to have isolated envs

PR #111, opened 2021-05-22
feat: core update and stamp picker

Cafe137:
`--stamp` options can be interactively picked from the currently available stamp list.

---

Will also include new `furious-commander` version which brings enhancements:

* Conflicting options are printed in help
* `noErrors` property for tolerating missing required options and arguments (for Stamp Picker)

### DIFF
--- a/src/service/stamp/index.ts
+++ b/src/service/stamp/index.ts
@@ -1,0 +1,18 @@
+import { Bee } from '@ethersphere/bee-js'
+import { exit } from 'process'
+import { CommandLog } from '../../command/root-command/command-log'
+
+export async function pickStamp(bee: Bee, console: CommandLog): Promise<string> {
+  const stamps = await bee.getAllPostageBatch()
+
+  if (!stamps.length) {
+    console.error('You need to have at least one stamp for this action.')
+    exit(1)
+  }
+
+  const choices = stamps.map(stamp => `${stamp.batchID} (${stamp.utilization})`)
+  const value = await console.promptList(choices, 'Please select a stamp for this action.')
+  const [hex] = value.split(' ')
+
+  return hex
+}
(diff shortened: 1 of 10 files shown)


Cafe137:
Will pass when new `furious-commander` is out: https://github.com/nugaon/furious-commander/pull/23

agazso:
I think it would be better if the list of stamps wouldn't "roll over" when you reach the end. At least currently (without labels) it's difficult to see how many stamps you have and therefore if you have seen them all. 

I was just scrolling indefinitely for a while until it was suspicious that I have too many stamps and I realised that I already reached the end and it started again from the beginning.

Or maybe adding a numbering would help too.

src/service/stamp/index.ts
```diff
@@ -0,0 +1,18 @@
+import { Bee } from '@ethersphere/bee-js'
+import { exit } from 'process'
+import { CommandLog } from '../../command/root-command/command-log'
+
+export async function pickStamp(bee: Bee, console: CommandLog): Promise<string> {
```
nugaon, line 5: please make a jsdoc what it returns exactly

nugaon (changes requested):
I had some problems when I tried out this PR:
- when I first saw the stamp picker, I didn't know what the the columns represent exactly. Maybe we would need a description before the choice printing how you can interpret the rows, like Choice: [stamp id] [utilization]
- The `--install-autocomplete` option is not printed in the help menu - that optionally should be only visible when I haven't installed yet.
- I had installed the auto complete as you mentioned but there were no output about its success or failiure.
- The autocomplete did not work for me after installation.

I used `bash`.

src/service/stamp/index.ts
Cafe137:
```
/**
 * Displays an interactive stamp picker to select a Stamp ID.
 *
 * A typical use case is to prompt the user for a stamp, when
 * the command requires one, but was not specified in `argv`.
 *
 * Makes a request via `bee` to fetch possible stamps.
 *
 * @returns {Promise<string>} Hex representation of the Stamp ID.
 */
```

nugaon (approved):
LGTM! we agreed on moving the autocompletion feature to other PR

PR #114, opened 2021-05-23
docs: monetary units

Cafe137:
Adds `PLUR` unit where applicable.

Also adds `wei` to `gas-price` option.

Adds 8 tests which check printed messages as well as help texts.


nugaon (approved):
LGTM!

PR #121, opened 2021-05-27
feat: bee-js update

Cafe137:
* Updates bee-js to 0.10.0
* Updates bee to 0.6.2
* Improves and shortens 404 error handling
* Also adds nicer 500 handling to some places
* Adds `pinning reupload`
* Adds `status`
* Renames `checkoutOne` to `cashoutOne`
* Adds `gasPrice` and `gasOption` to `cheque cashout`
* Adds a few more tests
* Updates `deposit` and `withdraw` return types

### DIFF
--- a/src/command/root-command/command-log.ts
+++ b/src/command/root-command/command-log.ts
@@ -1,6 +1,6 @@
 import { prompt } from 'inquirer'
 import { exit } from 'process'
-import { isNotFoundError } from '../../utils/error'
+import { isInternalServerError, isNotFoundError } from '../../utils/error'
 import { ErrorWithStatus } from '../../utils/types'
 import { Printer } from './printer'
 
@@ -131,7 +131,10 @@ export class CommandLog {
   }
 
   public printError(error: ErrorWithStatus, notFoundMessage: string): void {
-    if (isNotFoundError(error)) {
+    if (isInternalServerError(error)) {
+      this.error('Internal Server Error')
+      this.error('Check your Bee log to see what went wrong.')
+    } else if (isNotFoundError(error)) {
       this.error(notFoundMessage)
     } else {
       this.error(error.message)
--- a/src/command/root-command/command-log.ts
+++ b/src/command/root-command/command-log.ts
@@ -125,7 +125,7 @@ export class CommandLog {
   }
 
   public async promptList(choices: string[], message: string): Promise<string> {
-    const result = await prompt({ name: 'value', type: 'list', message, choices })
+    const result = await prompt({ name: 'value', type: 'list', message, choices, loop: false })
 
     return result.value
   }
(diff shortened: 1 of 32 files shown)


Cafe137:
@AuHau How did you make the reupload tests pass in bee-js? 😅

AuHau:
Heh, honestly I think I did not 😅 I thought with the latest PR I have figured out, but they are still very flaky :-( But your CI fails because of NPM not tests right?

src/command/root-command/command-log.ts
```diff
@@ -127,4 +129,15 @@ export class CommandLog {
 
     return result.value
   }
+
+  public printError(error: ErrorWithStatus, notFoundMessage: string): void {
```
nugaon, line 133: I suggest that you use `printBeeError` instead to clearly distinguish from other error messages.

nugaon: Also I would put notFoundMessage to an optional `options` object, and if `notFoundError` happens, use that one or the error message from the response.

src/utils/types/index.ts
```diff
@@ -8,3 +8,5 @@ export type BrandedType<T, N> = T & { __tag__: N }
 export type BrandedString<N> = BrandedType<string, N>
 
 export type FlavoredType<T, N> = T & { __tag__?: N }
+
+export type ErrorWithStatus = Error & { status?: number }
```
nugaon, line 12: You said next to me use `BeeError` instead

src/utils/error.ts
```diff
@@ -0,0 +1,9 @@
+import { ErrorWithStatus } from './types'
+
+export function isNotFoundError(error: ErrorWithStatus): boolean {
+  return error.status === 404 || error.message.includes('Not Found')
+}
+
+export function isInternalServerError(error: ErrorWithStatus): boolean {
+  return error.status === 500 || error.message.includes('Internal Server Error')
```
nugaon, line 8: the last part of the checking is not necessary.

src/utils/error.ts
```diff
@@ -0,0 +1,9 @@
+import { ErrorWithStatus } from './types'
+
+export function isNotFoundError(error: ErrorWithStatus): boolean {
+  return error.status === 404 || error.message.includes('Not Found')
```
nugaon, line 4: the last part of the checking is not necessary.

test/command/pinning.spec.ts
```diff
@@ -112,7 +112,7 @@ describe('Test Pinning command', () => {
     expect(consoleMessages[3]).toContain('Reuploaded successfully.')
   })
 
-  it('should allow reuploading pinned folder', async () => {
+  test.skip('should allow reuploading pinned folder', async () => {
```
nugaon, line 115: have we created an issue for the flaky reupload problem in the Bee repo? if so, please place a comment above this line with the issue link, starting with `FIXME` or create issue about it.

nugaon (approved):

PR #125, opened 2021-05-31
feat: status command

Cafe137:
image

### DIFF
--- a/src/command/status.ts
+++ b/src/command/status.ts
@@ -43,8 +43,13 @@ export class Status extends RootCommand implements LeafCommand {
 
   private async checkBeeVersionCompatibility(): Promise<void> {
     try {
-      await this.beeDebug.isSupportedVersion()
-      this.printSuccessfulCheck('Bee Version Compatibility')
+      const compatible = await this.beeDebug.isSupportedVersion()
+
+      if (compatible) {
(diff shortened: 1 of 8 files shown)


src/command/status.ts
```diff
@@ -0,0 +1,62 @@
...
+  private async checkBeeApiConnection(): Promise<void> {
+    try {
+      await this.bee.checkConnection()
+      this.printSuccessfulCheck('Bee API Connection')
+    } catch {
+      this.printFailedCheck('Bee API Connection')
+    }
+  }
+
+  private async checkBeeDebugApiConnection(): Promise<{ version: string }> {
```
nugaon, line 31: I don't really get why is it necessary to put it into an object instead of just returning a `string`?

nugaon (changes requested):
From the main concept I'm missing if my Bee node is not compatible why, and from which version it is.

Also, I tried out with Bee version 0.5.6 where all of my checkes remained satisfied even though it shouldn't be because of the postage stamps.
Actually I tried to upload something with it and I got the following:
```
$ swarm-cli upload ./README.md 
█ Failed to run command!

Not Found: Not Found
```

nugaon (approved):
LGTM, though I would suggest that it should print later the `supported version` on Bee compatibility failing.

PR #128, opened 2021-06-03
feat: add pss commands

Cafe137:
```
█ Available Commands:

pss send        Send to recipient or target with Postal Service for Swarm
pss receive     Receive message with Postal Service for Swarm
pss subscribe   Subscribe to messages with Postal Service for Swarm
```

### DIFF
--- a/src/command/pss/send.ts
+++ b/src/command/pss/send.ts
@@ -1,5 +1,7 @@
+import { readFileSync } from 'fs'
 import { LeafCommand, Option } from 'furious-commander'
 import { pickStamp } from '../../service/stamp'
+import { fileExists } from '../../utils'
 import { isHexString } from '../../utils/hex'
 import { PssCommand } from './pss-command'
 
@@ -25,16 +27,27 @@ export class Send extends PssCommand implements LeafCommand {
 
   @Option({
     key: 'data',
-    description: 'Message to be sent',
+    description: 'Text message to send',
     required: true,
+    conflicts: 'path',
   })
   public data!: string
+
+  @Option({
+    key: 'path',
+    description: 'Send raw data from file',
+    required: true,
+    conflicts: 'data',
+  })
+  public path!: string
 
   @Option({
     key: 'recipient',
     description: 'Public PSS key of the recipient',
   })
   public recipient!: string
+
+  message?: string | Uint8Array
 
   public async run(): Promise<void> {
     super.init()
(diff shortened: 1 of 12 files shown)


src/command/pss/send.ts
```diff
@@ -0,0 +1,81 @@
...
+import { LeafCommand, Option } from 'furious-commander'
+import { pickStamp } from '../../service/stamp'
+import { fileExists } from '../../utils'
+import { isHexString } from '../../utils/hex'
+import { PssCommand } from './pss-command'
+
+export class Send extends PssCommand implements LeafCommand {
+  public readonly name = 'send'
+
+  public readonly description = 'Send to recipient or target with Postal Service for Swarm'
```
nugaon, line 11: This description little bit confuses me. I can send message to a `target`, who is the `recipient` in the first interpretation for me. The meaning of `recepient` is somehow a public key here which changes how the message will be encrypted, but still the message will land on the `target` side. Moreover the `target` will be the mined message's prefix that will land at the _target node_ because of the kademlia routing.
So I would say here `Send a message to a target Bee node with Postage Service for Swarm`,
where the `target` is the target bee node's overlay address' prefix and the `recepient` is  the target Bee node's PSS public key to be more clear about these terms.

test/command/pss.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+    const invocation = invokeTestCli([
+      'pss',
+      'receive',
+      '--bee-api-url',
+      'http://localhost:11633',
+      '--timeout',
+      '10000',
+    ])
+    await sleep(1000)
+    await invokeTestCli(['pss', 'send', '--target', '00', '--data', 'Bzzz Bzzzz Bzzzz', ...getStampOption()])
```
nugaon, line 34: These PSS tests will work only in small network, because it is not clear which node will get the `00` prefixed message.
Anyhow, the `swarm-cli` tests only assure the usage of the exposed functionalities can happen as it was inteded.
So it is good as it is just a reminder if we may would change for a bigger test network where not every node sees any other ones.

nugaon (changes requested):

src/utils/index.ts
```diff
@@ -22,3 +25,25 @@ export function fileExists(path: string): boolean {
 export function isGateway(url: string): boolean {
   return url.includes('gateway.ethswarm.org')
 }
+
+export function enforceValidHexTopic(console: CommandLog, topic: string): void {
+  const hasCorrectLength = topic.startsWith('0x') ? topic.length === 66 : topic.length === 64
+  const hasCorrectPattern = new RegExp(/^(0x)?[a-f0-9]+$/g).test(topic)
```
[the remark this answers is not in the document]
agazso: This would make the regex case insensitive:
```suggestion
  const hasCorrectPattern = new RegExp(/^(0x)?[a-f0-9]+$/gi).test(topic)
```

test/command/pss.spec.ts
agazso: I cannot test now, but from the Bee code it seems to be possible to use empty string as a target too. Not sure how well the swarm-cli would handle that case :)

nugaon:
Also if I use pss subscribe, then I got 3 times the messages. maybe it happens because all nodes around the target pushes the same chunk?

nugaon (approved):
LGTM!

PR #147, opened 2021-06-11
refactor(test): describe command

nugaon:
Every test had the same init for console logs and config setups, this PR extract this into the `describeCommand` test utility function to maintain clearer test files.

### DIFF
--- a/test/command/cheque.spec.ts
+++ b/test/command/cheque.spec.ts
@@ -1,9 +1,7 @@
-import { existsSync, unlinkSync } from 'fs'
 import { Server } from 'http'
 import { bold, green } from 'kleur'
-import { join } from 'path'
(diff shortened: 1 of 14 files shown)


Cafe137 (approved):
Nice, these are actually very helpful changes, the tests look much nicer.

I have added a few changes as well - mostly using the utility functions which are now available instead of the `consoleMessages.length - 1` stuff.

Also fixed the test where we want to check if the config file has been created - it was actually correct, but very confusing.

Cafe137 (approved):
Nice, these are actually very helpful changes, the tests look much nicer.

I have added a few changes as well - mostly using the utility functions which are now available instead of the `consoleMessages.length - 1` stuff.

Also fixed the test where we want to check if the config file has been created - it was actually correct, but very confusing.

PR #152, opened 2021-06-14
docs: add automation examples

Cafe137:
[Preview the README](https://github.com/ethersphere/swarm-cli/blob/98dcb2cf0ad806d6e38653c75dad0d6f8c13f182/README.md)

- Fixes writing errors to stderr
- Sets exit code to `1` when `status` has failures
- Changes description tagline to match Swarm-CLI tagline
- Adds Automation section 
- Adds Table of Contents

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,7 +220,9 @@ swarm-cli upload -q README.md --stamp $STAMP
 ## Config
 
 The configuration file is placed in a hidden folder named `swarm-cli`.
+
 In case of Unix-based systems this config path will be: `$HOME/.swarm-cli`
+
 On Windows systems: `$HOME\AppData\swarm-cli`
 
 The configuration file is saved with `600` file permission.
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
 
 # Description
 
-> Command line interface tool for manage Bee node and utilize its functionalities
+> Manage your Bee node and interact with the Swarm network via the CLI
 
 The goal of this project is to handle most of the Swarm operations through CLI at some point in the future.
 For currently supported operations, see [Commands](##Commands) section.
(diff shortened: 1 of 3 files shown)


Cafe137:
@agazso @nugaon I am curious what you think about adding the Table of Contents. 

The README is quite long, I think it would be beneficial to break it up to smaller parts and/or utilise the wiki features.

But for now, adding a ToC gives a nicer overview of the topics which partially solves this problem.

README.md
```diff
@@ -9,9 +9,37 @@
...
+      * [From source](#from-source)
+   * [Usage](#usage)
+   * [Commands](#commands)
+   * [Example usage](#example-usage)
+   * [Usability Features](#usability-features)
+      * [Numerical Separator and Units](#numerical-separator-and-units)
+      * [Stamp Picker](#stamp-picker)
+      * [Identity Picker](#identity-picker)
+      * [Human Readable Topics](#human-readable-topics)
+      * [Automatizing tasks with Swarm-CLI](#automatizing-tasks-with-swarm-cli)
```
agazso, line 28:
```suggestion
      * [Automating tasks with Swarm-CLI](#automating-tasks-with-swarm-cli)
```

agazso:
> @agazso @nugaon I am curious what you think about adding the Table of Contents.
> 
> The README is quite long, I think it would be beneficial to break it up to smaller parts and/or utilise the Wiki feature.
> 
> But for now, adding a ToC gives a nicer overview of the topics which partially solves this problem.

I think the ToC is very nice. I have seen projects with even bigger ToC and I don't think that's a problem. In the case you would want to break it up into smaller parts I would suggest to put the documentation in another file (e.g. DOCUMENTATION.md) instead of using the wiki feature, because that way the cloned repo will also have access to the documentation.

Cafe137:
Thanks for the feedback @agazso, I agree the README structure and size is good for now.

What do you think of the examples and the writing? Is this good enough, or should we expand on it more?

agazso (commented):
`status` command seems to be missing from the Swarm-cli output `Available commands` section.

agazso:
> Thanks for the feedback @agazso, I agree the README structure and size is good for now.
> 
> What do you think of the examples and the writing? Is this good enough, or should we expand on it more?
> 
> Edit: I know we discussed an _"automation vs interactive usage"_ section, but it felt out of the place since there are already 2 sections about interactiveness, so I ended up creating one for the automation on its own.

I think this is an improvement already. The examples are useful, but I would be a bit cautious about using them because if we change the output of the (for example) stamp listing they will break but there is no automatic way to test those. So maybe it would be useful to provide examples that are less prone to breaking.

README.md
```diff
@@ -9,9 +9,37 @@
...
+   * [System environment](#system-environment)
+* [Development](#development)
+* [Contribute](#contribute)
+* [Maintainers](#maintainers)
+* [License](#license)
+
 # Description
 
-> Command line interface tool for manage Bee node and utilize its functionalities
+> Manage your Bee node and interact with the Swarm network via the CLI
```
agazso, line 42: Nice! After the merge let's change the about message of the repo to this.

agazso (approved):

PR #153, opened 2021-06-14
feat: add upload size prompt/warning/flag

Cafe137:
In `--verbose` mode, file or folder size is printed before uploading:

```
Upload size is approximately 2.00 megabytes
```

---

When upload size is too large:

```
? Folder size is larger than recommended value. Do you want to proceed? (Y/n) 
```

---

Fails in `--quiet` mode without explicit flag:

```
swarm-cli upload dist -q --stamp ...
Folder size is larger than recommended value.
Pass --size-check false to ignore this warning.
```

---

Needs to be disabled with `--size-check false`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,7 @@ dist
 
 # Dependency directory
 node_modules
+
+# Test data
+test/data/*
+!test/data/.gitkeep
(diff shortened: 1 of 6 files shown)


Cafe137:
Reviewable, but I need to write a few tests.

agazso (approved):

PR #167, opened 2021-06-18
feat: add --drop-name to upload, print stamp after usage, sync progress

Cafe137:
* Adds `--drop-name` flag to erase file name during upload
* Print stamp usage after `upload`, `feed print` and `feed upload`
* Change upload bar from `processed` to `synced` which works when reading file to memory

### DIFF
--- a/src/command/upload.ts
+++ b/src/command/upload.ts
@@ -52,7 +52,7 @@ export class Upload extends RootCommand implements LeafCommand {
   @Option({
     key: 'drop-name',
     type: 'boolean',
-    description: 'Erase file name when upliading a single file',
+    description: 'Erase file name when uploading a single file',
   })
   public dropName!: boolean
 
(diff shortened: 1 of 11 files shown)


src/command/upload.ts
```diff
@@ -156,8 +160,10 @@ export class Upload extends RootCommand implements LeafCommand {
     this.console.dim('Uploading was successful!')
     this.console.log(bold(`URL -> ${green(url)}`))
 
-    if (!this.usedFromOtherCommand) {
+    if (!usedFromOtherCommand) {
       this.console.quiet(this.hash)
+      printEnrichedStamp(await this.bee.getPostageBatch(this.stamp), this.console)
+      this.console.divider()
     }
   }
```
nugaon, line 168: I wouldn't print after a singular item a divider.

nugaon (approved):
https://github.com/ethersphere/swarm-cli/blob/6ad17096fd84266f4130c806be28f9a47bdbdf6a/src/service/stamp/index.ts#L35-L39

this is not the percentage, this is the normalization

PR #176, opened 2021-06-21
fix(cashout): set minimum of minimum to 1

attila-lendvai:
Otherwise transactions were created for cashing out zero valued cheques.

I couldn't really test this, so take it more as a suggestion than a well-tested PR.

### DIFF
--- a/src/command/cheque/cashout.ts
+++ b/src/command/cheque/cashout.ts
@@ -28,9 +28,9 @@ export class Cashout extends ChequeCommand implements LeafCommand {
     key: 'minimum',
     alias: 'm',
     type: 'bigint',
-    minimum: BigInt(0),
+    minimum: BigInt(1),
     description: 'Cashout cheques with balance above this value in PLUR',
-    default: BigInt(0),
+    default: BigInt(1),
   })
   public minimum!: bigint
 


src/command/cheque/cashout.ts
```diff
@@ -72,7 +72,8 @@ export class Cashout extends ChequeCommand implements LeafCommand {
     const cheques = await this.getFilteredCheques(this.minimum)
     this.console.info('Found ' + cheques.length + ' cheques.')
     for (const { amount, address } of cheques) {
-      await this.cashoutOne(address, amount)
+      if (amount > 0)
+        await this.cashoutOne(address, amount)
```
Cafe137, line 76: This check is not necessary.

By the time we get here in execution, the cheques are already filtered based on `minimum`, which is at least `1`.

Can you please revert this part so I can approve your other changes? Thank you.

attila-lendvai: i'm happy to revert this, but i thought that this makes the code more resilient, i.e. if something changes/breaks with the command line parsing part...

short of a response i'll update the PR later today.

attila-lendvai: @Cafe137 done.

Cafe137 (approved):
LGTM

PR #183, opened 2021-06-24
feat: add --encrypt to upload

Cafe137:
Adds `--encrypt` boolean option to `upload` command

### DIFF
--- a/test/command/upload.spec.ts
+++ b/test/command/upload.spec.ts
@@ -31,13 +31,13 @@ describeCommand('Test Upload command', ({ consoleMessages, hasMessageContaining 
     expect(command.hash?.length).toBe(64)
   })
 
-  it('should upload folder and encrypt', async () => {
+  it('should upload file and encrypt', async () => {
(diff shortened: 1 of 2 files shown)


test/command/upload.spec.ts
```diff
@@ -31,6 +31,18 @@ describeCommand('Test Upload command', ({ consoleMessages, hasMessageContaining
     expect(command.hash?.length).toBe(64)
   })
 
+  it('should upload file and encrypt', async () => {
```
nugaon, line 34: the output files should be checked at least whether they differ.

Cafe137: Hmm, if you do not pass the encryption key, then the content is 404.

If you do pass it, then you get back the original file.

I am not sure if this is worth testing.

Cafe137: https://github.com/ethersphere/bee/issues/2215

nugaon (approved):

PR #185, opened 2021-06-25
feat: add --curl global option

Cafe137:
Adds `--curl` global option using the new `bee-js` *Hooks* feature for debug purposes.

### DIFF
--- a/src/curl.ts
+++ b/src/curl.ts
@@ -16,7 +16,11 @@ const getDataString = (data?: unknown): string | null => {
     return '<stream>'
   }
 
-  return null
+  if (data instanceof Uint8Array) {
+    return '<buffer>'
+  }
+
+  return '<unknown>'
 }
 
 function printCurlCommand(request: BeeRequest): BeeRequest {
--- a/src/curl.ts
+++ b/src/curl.ts
@@ -1,4 +1,4 @@
-import { AxiosRequestConfig } from 'axios'
+import { BeeRequest, Utils } from '@ethersphere/bee-js'
 import { Readable } from 'stream'
 import { printer } from './printer'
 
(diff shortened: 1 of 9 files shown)


src/curl.ts
```diff
@@ -0,0 +1,40 @@
...
+  const methodString = request.method.toUpperCase() === 'GET' ? '' : ` -X ${request.method?.toUpperCase()}`
+  const dataString = request.data ? ` --data "${getDataString(request.data)}"` : ''
+  const command = `curl${methodString} ${request.url}${queryString} ${headers}${dataString}`
+  printer.print(command)
+
+  return request
+}
+
+export function registerCurlHook(): void {
+  if (typeof process.env.CURL_HOOK_ID === 'undefined') {
```
AuHau, line 37: Is this sort of "singleton" design? Only one hook would be registered?

src/curl.ts
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  if (data instanceof Uint8Array) {
+    return '<buffer>'
+  }
+
+  return '<unknown>'
+}
+
+function printCurlCommand(request: BeeRequest): BeeRequest {
```
AuHau, line 21: Damn, I am an idiot 🙄. The hook should not be required to return anything, I have just wrote wrong type for it...

src/curl.ts
Cafe137: Well, kind of.

The tests fail because all tests in a suite register the hook. ^^

I just went with an env variable so the `id` is not completely lost

src/curl.ts
Cafe137: No biggie. 🙂

src/curl.ts
AuHau: Ahhh, right. I was just wondering that for the normal swarm-cli life-cycle this does not make really sense, but sure for tests do :-) 👍

AuHau (commented):
LGTM!

AuHau (approved):

PR #188, opened 2021-07-01
fix: reduce spinner interval to avoid flickering on windows

Cafe137:


agazso (approved):
I tested it on windows and it does not flicker

PR #189, opened 2021-07-01
refactor: improve style and format consistency

Cafe137:


src/command/stamp/list.ts
```diff
@@ -62,6 +62,14 @@ export class List extends StampCommand implements LeafCommand {
...
-    orderedStamps.forEach(stamp => printStamp(stamp, this.console))
+
+    for (let i = 0; i < orderedStamps.length; i++) {
+      const stamp = orderedStamps[i]
+      printStamp(stamp, this.console)
+
+      if (i !== orderedStamps.length - 1) {
+        this.console.divider()
+      }
+    }
```
nugaon, line 73: we could create a similar function as `createKeyValue` just for listing with this divider.
For this particular case I could imagine passing the print hook function (`printStamp`) and the array whose elements needs to be divided.

nugaon (approved):
LGTM, I had one comment with which you can improve this concept.

PR #198, opened 2021-07-06
feat: add topology info to status command

Cafe137:


src/command/status.ts
```diff
@@ -61,6 +70,28 @@ export class Status extends RootCommand implements LeafCommand {
...
+      const { connected, population, depth } = await this.beeDebug.getTopology()
+
+      return {
+        connected,
+        population,
+        depth,
+      }
+    } catch {
+      return {
+        connected: 0,
```
agazso, line 88: This seems to return zeroes if it cannot connect to the debug API. However it's possible that there are many connections but the debug API is disabled. 

I think it would be better to not show the topology information when there is a problem with the debug API instead of showing possibly incorrect information.

src/command/status.ts
```diff
@@ -12,17 +12,30 @@ export class Status extends RootCommand implements LeafCommand {
...
-    this.console.divider()
     this.console.log(createKeyValue('Bee Version', version, 'Supported Version'.length))
     this.console.log(
       createKeyValue('Supported Version', SUPPORTED_BEE_VERSION + ' (' + SUPPORTED_BEE_VERSION_EXACT + ')'),
     )
     this.console.quiet('Bee version - ' + version)
     this.console.quiet('Supported version - ' + SUPPORTED_BEE_VERSION_EXACT)
+
+    if (topology) {
+      this.console.divider()
```
nugaon, line 31: Here, I would use different devider as this is a different section
```suggestion
      this.console.divider('=')
```

nugaon (approved):

PR #199, opened 2021-07-06
fix: handle payload size errors in pss send

Cafe137:


src/command/pss/send.ts
```diff
@@ -59,6 +60,16 @@ export class Send extends PssCommand implements LeafCommand {
       this.sendable = this.message
     }
 
+    const length = this.sendable.length
+
+    if (length > 4000) {
```
nugaon, line 65: this is not great, `this.spendable` can be either `string` or `Buffer`. As far as I know the default encoding is `UTF8` on `readFileSync` which returns back a `Buffer` instance, which is derived from the `Uint8Array` class. It means the string message can take 1 to 4 bytes per character, whose length won't be equal to the `Buffer`'s length.
You can try out this with the following snippet:
```js
const message = 'Árvíztűrő tükörfúrógép'
const buffer = Buffer.from(message)
console.log('string length', message.length)
console.log('buffer length', buffer.length)
```

nugaon (changes requested):
The problem came you haven't wrote test for the `UTF8` edge-cases.

nugaon (approved):
LGTM!

Issue #203, opened 2021-07-07
Prompt color system


Cafe137:
Use turquoise for prompt styling and related messages

PR #207, opened 2021-07-14
feat: establish color system with chalk

Cafe137:
- Changed green question mark to cyan in prompts (#203)
- Hexagon spinner is now orange
- Replaced `kleur` with `chalk`


nugaon (approved):
Cool!

PR #208, opened 2021-07-14
test: use correct pss addresses to avoid 'wants self' test errors

Cafe137:
The idea here is to extract the Worker PSS addresses to an environment variable, so tests can properly target the other node instead of just specifying `0x00` address prefix / target.

Also sets a different topic for each PSS test.

### DIFF
--- a/test/utility/address.ts
+++ b/test/utility/address.ts
@@ -2,7 +2,7 @@ import { invokeTestCli } from '.'
 import { Addresses } from '../../src/command/addresses'
 
 export async function getPssAddress(beeDebugApiUrl: string): Promise<string> {
-  const execution = await invokeTestCli(['stamp', 'addresses', '--bee-debug-api-url', beeDebugApiUrl])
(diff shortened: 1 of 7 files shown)


Cafe137:
This either does not fix all issues, or does not work at all. :/

test/utility/address.ts
```diff
@@ -0,0 +1,16 @@
+import { invokeTestCli } from '.'
+import { Addresses } from '../../src/command/addresses'
+
+export async function getPssAddress(beeDebugApiUrl: string): Promise<string> {
+  const execution = await invokeTestCli(['addresses', '--bee-debug-api-url', beeDebugApiUrl])
+
+  return (execution.runnable as Addresses).nodeAddresses.pssPublicKey
+}
+
+export function getWorkerPssAddress(stringLength: number): string {
```
agazso, line 10: Cool, it would be good to have this is bee-js too :)
https://github.com/ethersphere/bee-js/issues/250

agazso (approved):

PR #216, opened 2021-07-22
feat: set swarm orange as primary color

Cafe137:


Cafe137:
> code is fine, but about UX part I would like to hear @agazso's opinion too.

We already discussed this. The guideline is that only one element should have the brand orange color, so this PR will need to be reworked. It is super low priority though. 🙂

PR #217, opened 2021-07-22
feat: rework --skip-sync to --sync with connectivity check

Cafe137:
Removed `--skip-sync`, added `--sync`, and `--sync` is now opt-in

---

When debug API is not available:

image

---

When there are no connected peers:

image

---

Prompts have been removed, only a warning message is displayed.

### DIFF
--- a/src/command/upload.ts
+++ b/src/command/upload.ts
@@ -46,7 +46,7 @@ export class Upload extends RootCommand implements LeafCommand {
   @Option({
     key: 'sync',
     type: 'boolean',
-    description: 'Waiting for chunk synchronization over the network',
+    description: 'Wait for chunk synchronization over the network',
   })
   public sync!: boolean
 
--- a/src/command/upload.ts
+++ b/src/command/upload.ts
@@ -92,11 +92,11 @@ export class Upload extends RootCommand implements LeafCommand {
   public async run(usedFromOtherCommand = false): Promise<void> {
     this.initCommand()
 
-    await this.handleSyncSupport()
-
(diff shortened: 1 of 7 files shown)


AuHau:
LGTM, but I am wondering about the UX. Maybe I would be more proactive and just disabled the sync automatically without the prompt and only informed the user. The problem with this approach is that we would have to have a flag that could disable this behavior, which is not ideal. Although maybe we could use `--no-skip-sync`, I am not sure if the framework allows for bool flags to use `--no...` prefixes for `false` values.

I will leave the approval to @agazso as he is more aware of the UX side of swarm-cli.

Cafe137:
One thing I do see as an issue is that we are using the `--quiet` flag for non-interactive, but it probably should have its own option, as verbosity level should not effect interactiveness.

src/command/upload.ts
```diff
@@ -320,4 +319,32 @@ export class Upload extends RootCommand implements LeafCommand {
 
     return false
   }
+
+  private async handleSyncSupport(): Promise<void | never> {
+    if (this.quiet || !this.sync) {
```
agazso, line 324: This makes it not possible to turn on syncing in quiet mode?

Cafe137: This is only about printing the warnings. Sync will work regardless of verbosity level.

agazso: Then maybe it would be better to rename the function to `printWarningWhenNoConnectedPeers` or something more relevant.

nugaon (approved):
LGTM, althought I expected a successful run test also with `--sync`

PR #225, opened 2021-08-03
feat: add debug api correctness check when normal api is explicit and remote

Cafe137:


src/command/root-command/index.ts
```diff
@@ -71,4 +84,12 @@ export class RootCommand {
       }
     }
   }
+
+  private localRemoteUrlMismatch(): boolean {
+    return (
+      this.sourcemap['bee-api-url'] === 'explicit' &&
+      !this.beeApiUrl.startsWith('http://localhost:') &&
```
nugaon, line 91: Why did you add colon at the end of the string?

src/command/root-command/index.ts
```diff
@@ -71,4 +84,12 @@ export class RootCommand {
       }
     }
   }
+
+  private localRemoteUrlMismatch(): boolean {
```
nugaon, line 88: please add an other name to this function that reflects the role of this function and also add some jsdoc because it won't be straightforward later.

src/command/root-command/index.ts
nugaon: plus it won't be needed IMO. the two other condition may be enough

src/command/root-command/index.ts
```diff
@@ -31,7 +32,19 @@ export class RootCommand {
   public curl!: boolean
 
   public bee!: Bee
-  public beeDebug!: BeeDebug
+  public _beeDebug?: BeeDebug
+  public get beeDebug(): BeeDebug {
+    if (!this._beeDebug) {
+      this.console.error('Cannot ensure Debug API correctness!')
+      this.console.error('--bee-api-url is set explicitly to a non-local address.')
```
nugaon, line 39:
```suggestion
      this.console.error('--bee-api-url is set explicitly but --bee-debug-api-url.')
```

nugaon (changes requested):

nugaon (approved):
LGTM!

PR #237, opened 2021-09-01
feat: add manifest commands

Cafe137:


src/command/manifest/sync.ts
```diff
@@ -0,0 +1,68 @@
...
+    const forks = this.getValueForkMap(this.findAllValueForks(node))
+    for (const file of files) {
+      const fork = forks[file]
+
+      if (fork) {
+        fork.found = true
+        const remoteData = await this.bee.downloadData(Buffer.from(fork.node.getEntry).toString('hex'))
+        const localData = readFileSync(this.folder + '/' + file)
+
+        if (localData.equals(remoteData)) {
```
nugaon, line 46: it is quite heavy operation. That would be more performant _later when we have the chunker package_ is creating BMT hash of the local file and match it with the corresponding manifest node's entry, thereby you don't have to download the files into memory.

nugaon: so please add a TODO here about that

src/command/manifest/sync.ts
```diff
@@ -0,0 +1,68 @@
...
+          this.console.log('[ ok] ' + file)
+        } else {
+          const addition = await this.bee.uploadData(this.stamp, readFileSync(this.folder + '/' + file))
+          node.addFork(this.encodePath(file), Buffer.from(addition, 'hex') as Reference)
+          this.console.log('[upd] ' + file)
```
nugaon, line 51: maybe `up` would be more mathing as we use that abbreviation for the `upload` command as well.
```suggestion
          this.console.log('[ up] ' + file)
```

nugaon (changes requested):
please amend the readme as well

src/command/manifest/sync.ts
Cafe137: I am not sure if it is obvious. I am using three letters here to avoid upload and update ambiguity.

jest.config.ts
```diff
@@ -10,7 +10,9 @@ import { buyStamp } from './test/utility/stamp'
 export default async (): Promise<Config.InitialOptions> => {
   process.env.MAX_UPLOAD_SIZE = '5000000' // 5 megabytes
 
-  process.env.WORKER_PSS_ADDRESS = await getPssAddress('http://localhost:11635')
+  if (!process.env.SKIP_WORKER) {
```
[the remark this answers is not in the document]
Cafe137:
```
  /**
   * SKIP_WORKER can be enabled when running a subset of the tests manually,
   * which do not require any worker nodes, and therefor the stack
   * only consists a single queen node as well
   */
```

src/utils/index.ts
```diff
@@ -30,3 +41,41 @@ export function getByteSize(data: string | Uint8Array): number {
...
+}
+
+function removeLeadingDirectory(path: string, directory: string) {
+  directory = directory.startsWith('./') ? directory.slice(2) : directory
+  directory = directory.endsWith('/') ? directory : directory + '/'
+
+  return path.replace(directory, '')
+}
+
+export async function readdirDeepAsync(path: string, cwd: string): Promise<string[]> {
```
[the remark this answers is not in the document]
Cafe137: This is correct, follows Node.js naming: https://nodejs.org/api/fs.html#fs_fs_readdir_path_options_callback

src/utils/index.ts
```diff
@@ -30,3 +41,41 @@ export function getByteSize(data: string | Uint8Array): number {
...
+  const entries = []
+  for await (const entry of walkTreeAsync(path)) {
+    entries.push(cwd ? removeLeadingDirectory(entry, cwd) : entry)
```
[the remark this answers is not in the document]
Cafe137: My bad, it is an optional parameter

nugaon (approved):
LGTM, but please bump the mantaray-js version

PR #241, opened 2021-09-19
feat: improve error logging and troubleshooting

Cafe137:


nugaon (approved):
noice

Cafe137:
image

Cafe137:
image

src/utils/error.ts
```diff
@@ -1,9 +1,63 @@
...
+
+function isGenericErrorPattern(errorName: string, message: string | unknown): boolean {
+  if (!message || typeof message !== 'string') {
+    return true
+  }
+
+  errorName = errorName.toLowerCase()
+  message = message.toLowerCase()
+
+  return message === errorName || message === `${errorName}: ${errorName}`
```
nugaon, line 62: For me it is a bit vague what this function checks.

first, at the condition check if `message` is falsy or not string, then it is true.

and the last return I totally lost the tract why the `message` should correspond that format. can you write comments on these or a JSDoc please?

src/utils/error.ts
```diff
@@ -1,9 +1,63 @@
...
+    printer.printError('The command failed with error message: ' + error)
+  } else {
+    printer.printError('The command failed, but there is no error message available.')
+  }
+
+  if ((typeof error === 'string' && error) || message) {
+    printer.printError('')
+    printer.printError('There may be additional information in the Bee logs.')
+  } else {
+    printer.printError('')
```
nugaon, line 49: why don't you use `\n` instead?

src/utils/error.ts
```diff
@@ -1,9 +1,63 @@
...
+function isNotFoundError(error: unknown): boolean {
+  return hasStatusCode(error, 404)
+}
+
+function isInternalServerError(error: unknown): boolean {
+  return hasStatusCode(error, 500)
+}
+
+export function handleError(error: unknown, options?: BeeErrorOptions): void {
+  const message = getFieldOrNull(error, 'message')
```
nugaon, line 21: can't it be like this? as I checked you mainly using either this or that to print error message.
```suggestion
  const message = typeof error === 'string' ? error : getFieldOrNull(error, 'message')
```

PR #246, opened 2021-09-23
refactor: add bzz://<hash>/<path> support to manifest, improve listing

Cafe137:


src/utils/bzz-address.ts
```diff
@@ -0,0 +1,18 @@
...
+    if (url.startsWith('bzz://')) {
+      url = url.slice(6)
+    }
+    const parts = url.split('/')
+    this.hash = parts[0].toLowerCase()
+
+    if (!/[a-z0-9]{64,128}/i.test(this.hash)) {
+      throw new Error('Invalid BZZ address: expected 64 or 128 long hexadecimal hash')
+    }
+    const pathParts = parts.slice(1).filter(x => x)
```
nugaon, line 15: The mantaray structure allows `//` path in general, but on filesystem level and usage it would be better maybe to just throw an error instead.

src/utils/bzz-address.ts
```diff
@@ -0,0 +1,18 @@
...
+    this.path = pathParts.length ? pathParts.join('/') : null
```
nugaon, line 16: at this point we could use the operation system's file separator instead to construct `path`. This current setup may not work on Windows.

nugaon (changes requested):

src/command/manifest/manifest-command.ts
```diff
@@ -53,25 +87,91 @@ export class ManifestCommand extends RootCommand {
...
+    for (const fork of Object.values(node.forks)) {
+      const path = currentPath + Buffer.from(fork.prefix).toString('utf-8')
+
+      if (path.startsWith(prefix)) {
+        const reference = this.decodeReference(fork.node.getEntry as Reference)
+        const match = await this.getNodeForReference(reference)
+
+        return { node: match, prefix: path }
+      } else if (prefix.startsWith(path)) {
+        return this.findNodeForPrefix(this.decodeReference(fork.node.getEntry as Reference), prefix, path)
```
nugaon, line 171: the first parameter of `findNodeForPrefix` is already decoded in `reference` constant.

test/command/manifest.spec.ts
```diff
@@ -163,4 +195,60 @@ describeCommand('Test Upload command', ({ consoleMessages, hasMessageContaining
...
+    await invokeTestCli(['manifest', 'download', `${srcHash}/utils/types`, 'test/data/6'])
+    expect(consoleMessages[0]).toContain(['utils/types/config-option.ts'])
+    expect(consoleMessages[1]).toContain(['utils/types/index.ts'])
+  })
+
+  it('should download files in folder when specified fully with trailing slash', async () => {
+    await invokeTestCli(['manifest', 'download', `${srcHash}/utils/types/`, 'test/data/6'])
+    expect(consoleMessages[0]).toContain(['utils/types/config-option.ts'])
+    expect(consoleMessages[1]).toContain(['utils/types/index.ts'])
+  })
```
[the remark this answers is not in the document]
nugaon: and please add test case when I want to list or download files on a wrong, non-existing path.

nugaon (approved):

PR #264, opened 2021-10-08
fix: print relevant error message for bad manifest references

Cafe137:


nugaon:
on non-manifest swarm hash references I still get wrong error message

```
./node_modules/.bin/ts-node src/index.ts manifest list 82d3c5f5c757d852c04ae846ebedaf6723756a12770705d962ed249518705b2f
```

> The command failed with error message: Wrong reference length. Entry only can be 32 or 64 length in bytes
There may be additional information in the Bee logs.

PR #288, opened 2021-11-04
chore: revert package.json files

Cafe137:


AuHau (approved):
LGTM.

Btw. there is commit type `revert:` that I would maybe suggest to use instead of `chore:`

https://www.conventionalcommits.org/en/v1.0.0/#how-does-conventional-commits-handle-revert-commits

PR #293, opened 2021-11-08
chore: bump bee-js

AuHau:


nugaon (commented):
what happened with the `package-lock.json` file?

AuHau:
@nugaon good catch, I have been using old version of NPM that used the `version: 1` package-lock.json and hence the big diff. I fixed that.

nugaon (approved):

PR #305, opened 2021-12-19
feat: add feed download

Cafe137:


src/command/manifest/download.ts
```diff
@@ -13,6 +13,8 @@ export class Download extends ManifestCommand implements LeafCommand {
   public readonly name = 'download'
   public readonly description = 'Download manifest content to a folder'
 
+  public address!: BzzAddress
```
nugaon, line 16: well, it seems a bit interesting how you juggle with the `address` field between commands, but I suggest that you should define it in the `run` method as an optional parameter.

nugaon (commented):
please, write test for the feed resolution

PR #322, opened 2022-02-01
feat: add identity show command

Cafe137:


AuHau (approved):
LGTM, nice idea to separate displaying sensitive info only on demand! 👍

PR #371, opened 2022-04-25
fix: add cross-env and fix type error

Cafe137:
This does not solve all issues but fixes 2 things already. To be continued


Cafe137:
Superseded by https://github.com/ethersphere/swarm-cli/pull/391

Issue #355, opened 2022-04-08
feat: show TTL in human readable format


significance:
TTL: 11145595/60/60/24/7 seconds 😭

Cafe137:
The `swarm-cli stamp list` and `swarm-cli stamp show` commands print `TTL: 7231860 seconds`.

We have a `printStamp` function in `src/service/stamp/index.ts` which could be enhanced to convert this value to more comprehensible units.

agazso:
Closed by #395

PR #392, opened 2022-08-12
fix: convert seconds to human readable format issue #355

luissanchez0305:


Cafe137 (changes requested):
I have just tried it out and it is working nicely. 👍 

Example outputs:

```
TTL: 1 day, 8 hours, 53 minutes, 20 seconds
TTL: 80 days, 17 hours, 33 minutes, 30 seconds
```

I'd like to ask you to please install the `dbaeumer.vscode-eslint` extension to enable our ESLint configuration. It can be configured to fix style issues on save. I noticed the following changes need to be made, almost all of them are handled automatically:

 - `===` instead of `==`
 - `'` instead of `"`
 - No semicolons 
 - `const` instead of `var`
 - Empty line before a `return` statement
 - Whitespace around symbols

This is usually caught by the CI itself but since this is a fork as a temporary workaround, the checks are not running.

One more thing I'd like to ask is to move the function `secondsToDhms` to `utils/index.ts` and export it from there since it is a generic utility function.

luissanchez0305:
Will create a new PR since this is not relevant now that I have access to the swarm-cli rep

PR #395, opened 2022-08-15
fix: convert TTL to human readable format

luissanchez0305:
Referring issue
#355 

image

Added a util method to convert seconds to time human-readable format DHMS


Cafe137 (approved):
Nice! Feel free to merge

Issue #379, opened 2022-06-04
feat: display account balance in BZZ


significance:
it is not possible to determine the account balance in BZZ, only the chequebook balance.

please add the account balance for completeness. 🙇

Cafe137:
The `swarm-cli cheque balance` command prints the chequebook xBZZ balance in PLUR.

I propose to change this command to be a top-level `swarm-cli balance` command, which not only prints the chequebook balance, but also the node balances. This wallet endpoint is a new addition in Bee and is already supported in `bee-js@5.0.0`.

After merging #391 this will be available on the BeeDebug object: `this.beeDebug.getWalletBalance()`

agazso:
Fixed by #400

PR #396, opened 2022-08-15
fix: convert PLUR to BZZ on cheque balance

luissanchez0305:
Referring issue 
#379 

image

Used the conversion rate of 1BZZ = 10e16 PLUR on the display of the cheque balances command


Cafe137 (changes requested):
There are two different balances in Bee:

- one belongs to the node
- one belongs to the chequebook

---

Node: `http://localhost:1635/wallet`

Example response:

```
{
  "bzz": "1000000000000000000",
  "xDai": "1000000000000000000",
  "chainID": 0,
  "contractAddress": "36b7efd913ca4cf880b8eeac5093fa27b0825906"
}
```

---

Chequebook: `http://localhost:1635/chequebook/balance`

Example response: 

```
{
  "totalBalance": "1000000000000000000",
  "availableBalance": "1000000000000000000"
}
```

---

The chequebook balance is already implemented. The task at hand is to also add the node balance to the output of this command. In Swarm-CLI, all the commands are classes which inherit the Bee and BeeDebug instances (bee-js library). To query the wallet, use the BeeDebug: `this.beeDebug.getWalletBalance()`

---

Reaction to your changes: it makes sense to display the balance in BZZ instead of PLUR what we have now, so let's keep it.

However, math with the JS Number type is imprecise for these calculations, I suggest using `bignumber.js`. I suggest copying over this class from `bee-dashboard` and keep the relevant parts for printing significant digits: https://github.com/ethersphere/bee-dashboard/blob/master/src/models/Token.ts

Issue #380, opened 2022-06-04
feat: predict how much a stamp will cost


significance:
it would be great to know how much a stamp will cost, before buying it.

Cafe137:
When buying a stamp with the `swarm-cli stamp buy` command, the `amount` and `depth` are required fields which are passed by the user. These are used in the formula for calculating the postage batch price.

Calculate and print this BZZ cost, which is already implemented in Bee Dashboard and can be copied here: https://github.com/ethersphere/bee-dashboard/blob/master/src/utils/index.ts#L218

When running without the `--yes` flag, prompt the user if they want to continue the purchase with the given price.

PR #402, opened 2022-08-19
feat: calculate and show stamp cost

luissanchez0305:
Reference issue
#380 

image

image

### DIFF
--- a/src/command/stamp/buy.ts
+++ b/src/command/stamp/buy.ts
@@ -47,6 +47,8 @@ export class Buy extends StampCommand implements LeafCommand {
   @Option({ key: 'wait-usable', description: 'Wait until the postage stamp becomes usable', type: 'boolean' })
   public waitUsable!: boolean
 
+  public estimatedConfirm!: boolean
+
   // CLASS FIELDS
 
   public postageBatchId!: string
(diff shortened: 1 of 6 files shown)


agazso:
While this is technically correct, but the usability of the values are quite bad. For example in the screenshots it is very difficulty to tell what is the exact value, because there are so many leading zeroes.

Can we improve that either by using scientific notation as well or use some clever formatting to make it easier to read? Possibly in a different PR.

Cafe137:
We have a top level option in swarm-cli: `--yes`

This flag is for accepting all prompts, such as the price confirmation in the buy command.

Command classes can access this value as follows: `this.yes`

Can you please drop the `--no-estimate-confirm` flag in favour of the existing `--yes` option? Your approach is also correct and we used to have custom flags for each of these prompts, but we ended up consolidating them with the `--yes` flag for better UX and simpler code.

Cafe137:
Also please change the PR title to comply with the format.

Something like `feat: predict stamp cost`. Or maybe reword `predict` to `display`, `confirm` or `prompt`, your call.

Cafe137:
```
  toSignificantDigits(digits = 4): string {
    const asString = this.toDecimal.toFixed(this.decimals)

    let indexOfSignificantDigit = -1
    let reachedDecimalPoint = false

    for (let i = 0; i < asString.length; i++) {
      const char = asString[i]

      if (char === '.') {
        reachedDecimalPoint = true
        indexOfSignificantDigit = i + 1
      } else if (reachedDecimalPoint && char !== '0') {
        indexOfSignificantDigit = i
        break
      }
    }

    return asString.slice(0, indexOfSignificantDigit + digits)
  }
```

Taken from Token.ts: https://github.com/ethersphere/bee-dashboard/blob/master/src/models/Token.ts

src/command/stamp/buy.ts
```diff
@@ -62,28 +62,39 @@ export class Buy extends StampCommand implements LeafCommand {
       this.waitUsable = await this.console.confirm('Would you like to enable it now?')
     }
 
-    const spinner = createSpinner('Buying postage stamp. This may take a while.')
+    const _estimated = BigNumber(Number(this.amount)).multipliedBy(BigNumber(2).pow(this.depth))
+    const PLURConvertionRate = BigNumber(10).pow(16)
+    const estimated = BigNumber(_estimated).dividedBy(PLURConvertionRate)
```
Cafe137, line 67: The duplication of `estimated` and `_estimated` is not necessary here.

I suggest first creating `const PLURConvertionRate = BigNumber(10).pow(16)` and then you can compute `estimated` in one go. One side note is that `estimated` does not tell you what it estimates, so I think `estimatedCost` would be a better name.

Careful with the `Number(this.amount)`, since `this.amount` is already a `bigint`, converting it back to `Number` will make it lose precision.

So I think it could look like this:

```
const estimatedCost = BigNumber(this.amount).multipliedBy(BigNumber(2).pow(this.depth)).dividedBy(PLURConvertionRate) // in bzz
```

src/utils/index.ts
```diff
@@ -194,3 +194,26 @@ export function isPrivateKey(string: string): boolean {
...
+
+    if (e > 20) {
+      e -= 20
+      x /= Math.pow(10, e)
+      res += new Array(e + 1).join('0')
+    }
+  }
+
+  return res
+}
```
Cafe137, line 219: I am not sure if this function is working. I was testing it like this (proper unit tests would be better):

```
console.log(toFixed(1000))
console.log(toFixed(10))
console.log(toFixed(1))
console.log(toFixed(0.1))
console.log(toFixed(0.01))
console.log(toFixed(0.000001))
console.log(toFixed(0.000000000000001))
```

Output:

```
0
0
0
0
0
0
0.000000000000001
```

luissanchez0305: Yes, it was intended solely for decimals. I completely removed it

src/command/stamp/buy.ts
luissanchez0305: The problem with this `BigNumber(this.amount)` is that this.amount is a bigint, and is not being accepted by BigNumber as a valid parameter type, that's why I cast it to `number`
image

agazso: In that case it would be better to convert `this.amount` to string so that it does not lose precision.

Cafe137:
When running `swarm-cli` in `--quiet` mode the estimated cost is not printed but the prompt appears:

```
stamp buy --amount 596046400 --depth 24 -q
? Please confirm if you agree (Y/n)
```

src/utils/index.ts
Cafe137: Nice, thank you.

If `toFixed` is dead code please remove it

Cafe137:
Regarding the `fix: added the quiet prompt to stamp buy logic` commit

Passing `--quiet` without `--yes` simply terminates execution with exit code `0` but the purchase does not actually happen.

It would be better to allow running the command in `--quiet` without having to pass `--yes`.

Another solution would be throwing an error in this case with a message that explains `"This operation costs BZZ, pass the --yes option to allow it"`, but the other solution is better IMO.

Issue #361, opened 2022-04-08
feat: show TTL and exclude expired stamps in choose a stamp dialogue


significance:
```
? Please select a stamp for this action (Use arrow keys)
❯ 54a68fe89e85a28891e99f7f11b3ada55a6641d5766cb2173f3375fd25260c3e (25%) expires in 5 days
  99e78dab364aa305f4cc3af390ca01f7f03d0612da7276b4e225486d1af46284 (0%)   expires in 2 days
  eba0540f00fbd2641176b50313c51409073eba520ecac4748aecb2187bfb14e5 (0%)   ...
  4270e025e3a28985eebb6c5388bdaf14cb79f8adbecdfb9a311c73b4f86b174b (0%)
  83ec017b0d756e2ffcb274bb32eff9e8183587e64bb44e5b34c16ffce302c6b7 (100%)
  55b7d8b458b9419dba53be2ce337e4f8d4684a39e73a10b575ed22becf34209a (100%)
  2efb8717b7daeef25763005a3900799e9203af765ef8874e87d6ae887ee94e06 (50%)
  ```

PR #403, opened 2022-08-23
feat: improve stamp ttl display and exclude expired stamps

luissanchez0305:
Reference issue #361 

image

### DIFF
--- a/src/service/stamp/index.ts
+++ b/src/service/stamp/index.ts
@@ -19,7 +19,7 @@ export async function pickStamp(beeDebug: BeeDebug, console: CommandLog): Promis
   const stamps = ((await beeDebug.getAllPostageBatch()) || []).map(enrichStamp)
 
   const choices = stamps
-    .filter(stamp => stamp.batchTTL > 0)
+    .filter(stamp => stamp.usable || stamp.batchTTL > 0)
     .map(stamp => `${stamp.batchID} (${stamp.usageText}) expires in ${secondsToDhms(stamp.batchTTL, true)}`)
 
   if (!choices.length) {
--- a/src/service/stamp/index.ts
+++ b/src/service/stamp/index.ts
@@ -25,7 +25,7 @@ export async function pickStamp(beeDebug: BeeDebug, console: CommandLog): Promis
 
   const filterChoices = stamps.filter(stamp => stamp.batchTTL > 0)
   const choices = filterChoices.map(
-    stamp => `${stamp.batchID} (${stamp.usageText}) expires in ${secondsToDhms(stamp.batchTTL, false)}`,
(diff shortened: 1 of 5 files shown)


src/service/stamp/index.ts
```diff
@@ -23,7 +23,10 @@ export async function pickStamp(beeDebug: BeeDebug, console: CommandLog): Promis
     exit(1)
   }
 
-  const choices = stamps.map(stamp => `${stamp.batchID} (${stamp.usageText})`)
+  const filterChoices = stamps.filter(stamp => stamp.batchTTL > 0)
+  const choices = filterChoices.map(
```
Cafe137, line 27: I'd rather not create a `filterChoices` variable, instead just chain the functionals:

```
const choices = stamps.filter(x => x.batchTTL > 0)
    .map(/* ... */)
```

But if you want to keep it as a separate variable I'd suggest a better name such as `activeStamps`

src/utils/index.ts
```diff
@@ -51,18 +52,26 @@ export function getByteSize(data: string | Uint8Array): number {
...
-  return dDisplay + hDisplay + mDisplay + sDisplay
+  return resumed
+    ? dDisplay.length
+      ? dDisplay
+      : hDisplay.length
+      ? hDisplay
+      : mDisplay.length
+      ? mDisplay
+      : sDisplay
+    : dDisplay + hDisplay + mDisplay + sDisplay
```
Cafe137, line 74: I am not sure about the name `resumed`, it does not tell me what it means without reading the code. Maybe call it `shortFormat`?

I propose this structure to avoid the nested ternaries which are a bit difficult to read at first

```
if (shortFormat) {
    // print only the most significant value
    return dDisplay || hDisplay || mDisplay || sDisplay
}

return dDisplay + hDisplay + mDisplay + sDisplay
```

src/service/stamp/index.ts
```diff
@@ -24,19 +24,14 @@ export async function pickStamp(beeDebug: BeeDebug, console: CommandLog): Promis
     exit(1)
   }
 
-  const filterChoices = stamps.filter(stamp => stamp.batchTTL > 0)
-  const choices = filterChoices.map(
-    stamp => `${stamp.batchID} (${stamp.usageText}) expires in ${secondsToDhms(stamp.batchTTL, true)}`,
-  )
+  const choices = stamps
+    .filter(stamp => stamp.batchTTL > 0)
+    .map(stamp => `${stamp.batchID} (${stamp.usageText}) expires in ${secondsToDhms(stamp.batchTTL, true)}`)
```
Cafe137, line 29: We have this check:

```
if (!stamps.length) {
    console.error('You need to have at least one stamp for this action.')
    exit(1)
}
```

It should come after this part, and with `stamps` replaced with `choices`. Because right now it is possible that the user only has expired stamps and they are prompted with an empty list.

Issue #375, opened 2022-04-29
fix: i incorrectly thought it was truncat...


significance:
maybe remove the ellipsis

```
root@sig-bee-main-02:~/bee# swarm-cli feed print -v --stamp 2c6ea071030ea672a5502472b0acdb70aafa18cb7eda1bf142b2948241122153 -i main
Looking up feed topic 0000000000000000000000000000000000000000000000000000000000000000...
Chunk Reference: d755aa1e5b2e059f7013dbdd6e1134fbc00976c5254c176a74ef24f1ce60f06f
Chunk Reference URL: http://localhost:1633/bzz/d755aa1e5b2e059f7013dbdd6e1134fbc00976c5254c176a74ef24f1ce60f06f/
Feed Index: 0000000000000005
Next Index: 0000000000000006
Feed Manifest: 99b471e54c25c5fa51bb698be28d04ebe0313fb5ca7dbb946da01e4dea141e15
Feed Manifest URL: http://localhost:1633/bzz/99b471e54c25c5fa51bb698be28d04ebe0313fb5ca7dbb946da01e4dea141e15/
Number of Updates: 6
```

PR #404, opened 2022-08-23
fix: spinner on feed print command

luissanchez0305:
Reference issue #375 

image

### DIFF
--- a/src/command/feed/print.ts
+++ b/src/command/feed/print.ts
@@ -5,7 +5,9 @@ import { isSimpleWallet, isV3Wallet } from '../../service/identity'
 import { Identity } from '../../service/identity/types'
 import { pickStamp } from '../../service/stamp'
 import { createKeyValue } from '../../utils/text'
+import { createSpinner } from '../../utils/spinner'
 import { FeedCommand } from './feed-command'
+import { ExecException } from 'child_process'
 
 export class Print extends FeedCommand implements LeafCommand {
   public readonly name = 'print'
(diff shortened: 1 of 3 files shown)


src/command/feed/print.ts
```diff
@@ -26,26 +27,39 @@ export class Print extends FeedCommand implements LeafCommand {
...
+      this.console.verbose(createKeyValue('Chunk Reference URL', `${this.bee.url}/bzz/${reference}/`))
+      this.console.verbose(createKeyValue('Feed Index', feedIndex))
+      this.console.verbose(createKeyValue('Next Index', feedIndexNext))
+      this.console.verbose(createKeyValue('Feed Manifest', manifest))
+
+      this.console.quiet(manifest)
+      this.console.log(createKeyValue('Feed Manifest URL', `${this.bee.url}/bzz/${manifest}/`))
+      this.console.log(createKeyValue('Number of Updates', parseInt(feedIndex, 10) + 1))
+    } catch (ex: any) {
+      spinner.stop()
```
Cafe137, line 54: This can be removed since it is already in the `finally` block

src/command/feed/print.ts
```diff
@@ -26,26 +27,39 @@ export class Print extends FeedCommand implements LeafCommand {
...
+      this.console.info('Feed topic lookup error:')
+      this.console.error(`Status: ${ex.response.status} Message: ${ex.response.statusText}`)
+      this.console.error(ex.message)
+    } finally {
+      if (spinner.isSpinning) {
+        spinner.stop()
+      }
```
Cafe137, line 61: No need to wrap it in the condition, you can keep it a one liner with only `spinner.stop()`

src/command/feed/print.ts
```diff
@@ -26,26 +27,39 @@ export class Print extends FeedCommand implements LeafCommand {
...
+      this.console.verbose(createKeyValue('Next Index', feedIndexNext))
+      this.console.verbose(createKeyValue('Feed Manifest', manifest))
+
+      this.console.quiet(manifest)
+      this.console.log(createKeyValue('Feed Manifest URL', `${this.bee.url}/bzz/${manifest}/`))
+      this.console.log(createKeyValue('Number of Updates', parseInt(feedIndex, 10) + 1))
+    } catch (ex: any) {
+      spinner.stop()
+      this.console.info('Feed topic lookup error:')
+      this.console.error(`Status: ${ex.response.status} Message: ${ex.response.statusText}`)
```
Cafe137, line 56: This line throws because response does not always exist on `ex`.

e.g.: `ERROR Cannot read properties of undefined (reading 'status')`

src/command/feed/print.ts
```diff
@@ -26,26 +27,39 @@ export class Print extends FeedCommand implements LeafCommand {
...
-    const { reference, feedIndex, feedIndexNext } = await reader.download()
+    const spinner = createSpinner('Looking up feed topic ' + topic)
+    spinner.start()
+    try {
+      const addressString = this.address || (await this.getAddressString())
+      const reader = this.bee.makeFeedReader('sequence', topic, addressString)
+      const { reference, feedIndex, feedIndexNext } = await reader.download()
+
+      if (!this.stamp) {
+        this.stamp = await pickStamp(this.beeDebug, this.console)
```
Cafe137, line 38: `spinner` needs to be stopped before this and then resumed, otherwise both the spinner and the stamp picker are active and it is glitchy

Cafe137:
It seems the function `getAddressStringFromIdentity` has a bug.

In the `if (isSimpleWallet(wallet, identityType))` branch, the private key must not start with `0x`.

Here is a simple fix which will probably fix the tests as well

```
const privateKey = wallet.privateKey.replace('0x', '')
const buffer = Buffer.from(privateKey, 'hex')
const ethereumWallet = Wallet.fromPrivateKey(buffer)
```

Cafe137:
Lastly please change the PR title to use imperative mood: e.g. `fix: add spinner to feed print command`

src/command/feed/print.ts
```diff
@@ -26,26 +27,39 @@ export class Print extends FeedCommand implements LeafCommand {
...
+      this.console.verbose(createKeyValue('Next Index', feedIndexNext))
+      this.console.verbose(createKeyValue('Feed Manifest', manifest))
+
+      this.console.quiet(manifest)
+      this.console.log(createKeyValue('Feed Manifest URL', `${this.bee.url}/bzz/${manifest}/`))
+      this.console.log(createKeyValue('Number of Updates', parseInt(feedIndex, 10) + 1))
+    } catch (ex: any) {
+      spinner.stop()
+      this.console.info('Feed topic lookup error:')
+      this.console.error(`Status: ${ex.response.status} Message: ${ex.response.statusText}`)
```
Cafe137, line 58: Can we do something about this line please?
Maybe only execute it if `'response' in ex`

PR #411, opened 2022-08-28
chore: change chunk to root chunk

Cafe137:


agazso:
Let's discuss this, because the change introduces a new, inconsistent naming convention.

I think the best would be calling them simply `references`, because that is inline with the Bee documentation and also does not allow misinterpretation as would calling them `root hashes`.

PR #431, opened 2022-11-25
feat: staking support

### LINKED ISSUE #422: Add staking support
Since Bee version 1.9 there is support for staking with the /stake endpoint. It would be good if swarm-cli also supported that.

Depends on: https://github.com/ethersphere/bee-js/issues/786

AuHau:
Closes #422


src/command/balance.ts
```diff
@@ -16,24 +17,26 @@ export class Balance extends RootCommand implements LeafCommand {
     await super.init()
 
     this.console.dim('Looking up balance...')
+    const stakeBzz = await this._beeDebug.getStake()
```
Cafe137, line 20: IMO you can use `beeDebug` instead of `_beeDebug`

The one with underscore prefix does not do a hard requirement for debug api availability, but in this command we need it.

Cafe137 (commented):
The usage of the stake command is a bit different in philosophy than others.

Following the current principles there would be a `staking` group with `show` and `deposit` subcommands, ie. read and write are completely separated.

Even though I prefer your consolidated approach, I see some possible improvements.

I'd move the `Be aware, depositing stake is irreversible! It can not be withdrawn later on.` message to a `Y/n` prompt in `stake`, see `buy.ts` price confirmation. We have top level `--yes` flag for this and is affected by `--quiet` mode as well.

On the first deposit, does it have to be exactly 10 BZZ or gte 10 BZZ? In either case, I'd add a helper for this so the CLI tries to correct you. IMO this is better than documenting in the command's description. Having too much info there breaks the pattern a bit 🙂 
image

Lastly, to make it super obvious, and not sure about this, but when simply running the `stake` command in non-quiet mode I'd also print a helper that you can deposit with the `--deposit` flag. I can see a little confusion when running the `stake` command, not reading the `--help`, and being confused where to put the `100_000T`

Cafe137 (approved):
Awesome 🙂

PR #464, opened 2023-08-14
feat: add utility lock and unlock commands

Cafe137:


attila-lendvai:
@Cafe137 this works great, thanks!

unfortunately, i have just realized that this only works for the blockchain key, which is enough when someone wants to mine a new node into a neighborhood. but just now, when i wanted to migrate a node from one environment to another, i realized that the other keys (`libp2p_v2.key` and `pss.key`) are also encrypted with the password, so i cannot migrate the node.

i'll probably proceed by changing my service code to assign a separate password for each node it starts, but i wanted to let you know about this uncovered use-case, too.

Issue #479, opened 2023-10-26
When purchasing stamps, include effective volume as well


NoahMaizels:
Include effective volume from the table [here](https://docs.ethswarm.org/docs/learn/technology/contracts/postage-stamp#effective-utilisation-table) so users have a better idea of how much they can safely upload

Cafe137:
This can be implemented in stamp/buy.ts:

```
const estimatedCost = Utils.getStampCostInBzz(this.depth, Number(this.amount))
const estimatedCapacity = new Storage(Utils.getStampMaximumCapacityBytes(this.depth))
const estimatedTtl = Utils.getStampTtlSeconds(Number(this.amount))

this.console.log(createKeyValue('Estimated cost', `${estimatedCost.toFixed(3)} BZZ`))
this.console.log(createKeyValue('Estimated capacity', estimatedCapacity.toString()))
this.console.log(createKeyValue('Estimated TTL', secondsToDhms(estimatedTtl)))
```

An additional utility function is required which calculates effective volume based on the table, we don't have that anywhere yet.

PR #485, opened 2023-12-05
feat: display estimated effective volume during buy (#479)

LevilkTheReal:
Issue: https://github.com/ethersphere/swarm-cli/issues/479

Solution:
Screenshot 2023-12-05 at 15 34 12

@Cafe137 FYI

I could not run the tests. I am getting this error. Any idea?
Screenshot 2023-12-05 at 16 34 37

(I've extended the tests with this line tho: `expect(consoleMessages[2]).toBe('Estimated effective volume: 41.17 GB')`)

Another thing. The Storage class is not that precise. For instance, it converts the incoming effective bytes to 41.17 GB which should be 44.21 GB. 

Also, there are different converters/repos. Is it worth creating a shared one?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## [2.8.0](https://www.github.com/ethersphere/swarm-cli/compare/v2.7.0...v2.8.0) (2024-02-08)
+
+
+### Features
+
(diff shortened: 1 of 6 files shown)


Cafe137:
@LevilkTheReal For depths <= 21, I would maybe remove the message:

```
Estimated effective volume: 0 B
```

It looks like a bug.

Instead I would print a human friendly message, something like "effective volume is likely to be smaller than estimated capacity".

What do you think?

PR #490, opened 2024-01-25
feat: manifest sync improvement (#489)

### LINKED ISSUE #489: Optimise `manifest sync` by comparing hash obtained via bmt-js instead of downloading file
We already have bmt-js installed in the project.

```
src/command/manifest/sync.ts
```
vvv

```
const remoteData = await this.bee.downloadData(Buffer.from(fork.node.getEntry).toString('hex'))
```

This is the offender line that can be optimised.

tomicvladan:
The manifest sync command uses bmt-js to check changes instead of downloading all files.

Closes #489


tamas6:
> The manifest sync command uses bmt-js to check changes instead of downloading all files.
> 
> Closes #489

Long awaited

PR #501, opened 2024-05-23
Adds capability for the feed command, to specify index

ottpeter:
Adds capability for the feed command, to specify index, when doing read or write.

### DIFF
--- a/test/command/feed.spec.ts
+++ b/test/command/feed.spec.ts
@@ -110,6 +110,9 @@ describeCommand(
       expect(getLastMessage()).toContain('Number of Updates')
       expect(getLastMessage()).toContain('2')
     })
+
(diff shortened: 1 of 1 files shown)


src/command/feed/feed-command.ts
```diff
@@ -38,6 +38,9 @@ export class FeedCommand extends RootCommand {
   @Option({ key: 'password', alias: 'P', description: 'Password for the wallet' })
   public password!: string
 
+  @Option({ key: 'index', description: 'Index for the feed to write to, or read from', required: false })
```
aranyia, line 41: Maybe it could simply say "Feed index to write to or read from"?

ottpeter: Good idea. I'll make this change.

PR #569, opened 2025-03-10
feat: add act commands

ferencsarai:


test/command/download.spec.ts
```diff
@@ -4,18 +4,22 @@ import { join } from 'node:path'
...
-    const invocation = await invokeTestCli(['upload', 'test/message.txt', ...getStampOption()])
+    const file = 'message.txt'
+    const invocation = await invokeTestCli(['upload', 'test/' + file, ...getStampOption()])
     const hash = (invocation.runnable as Upload).result.getOrThrow()
     consoleMessages.length = 0
     await invokeTestCli(['download', hash.toHex(), '--stdout'])
-    expect(consoleMessages[0]).toContain('Hello Swarm!')
+    expect(consoleMessages[0]).toContain('OK')
+    const fileContent = readFileSync(join(hash.toHex(), file), 'utf-8')
+    expect(fileContent).toContain('Hello Swarm!')
```
Cafe137, line 22: I am not sure about this change.

With `--stdout`, it should NOT print the hash, just the content. So the previous version seems correct, this one does not?

Can you please check why this was needed?

ferencsarai: Yes, this is wrong.

PR #679, opened 2026-04-20
feat: implement upload history

### LINKED ISSUE #486: Save hash with postage stamp in json file
Currently, when we upload new data there is no record of our data and how it connects to stamp.

Would be nice if there is a simple JSON file that saves this when we upload.

So it would be JSON with stamp hash as the main key and under it, we should nest all the IDs of hashes that are connected with this stamp and ideally its some kind of Label:Hash pair so something like this should be our solution.

```
{
    "hashID_12345": {
        "Label1": "DataHash_A",
        "Label2": "DataHash_B",
        "Label3": "DataHash_C"
    },
    "hashID_67890": {
        "Label1": "DataHash_D",
        "Label2": "DataHash_E",
        "Label3": "DataHash_F"
    },
    "hashID_24680": {
        "Label1": "DataHash_G",
        "Label2": "DataHash_H",
        "Label3": "DataHash_I"
    }
}
```

### LINKED ISSUE #678: Implement upload history
# Problem statement

Upload commands addressed via `swarm-cli upload` and `swarm-cli feed upload` print the resulting address (also called "reference" or "hash"), but it is not persisted anywhere. If the user does not manually copy the hash and save it elsewhere, it will be lost (just the link to it, not the actual data on the network).

# New feature

## Upload history

Implement a new feature in `swarm-cli` that maintains an upload history in a JSON file. A similar system is already implemented for the `identity` commands, see `src/command/root-command/command-config.ts`.

The two upload commands are implemented in `src/command/upload.ts` and `src/command/feed/upload.ts`. In the former, the resulting hash is tracked in the class field `public result: Optional<Reference> = Optional.empty()`, and in the latter it is also in a class field `public feedManifest?: Reference`. After successful uploads, store the Swarm hash from these fields in the history and persist.

## `history list` command

Implement a new group (`history`) and a new command (`list`) that prints the upload history in a nice, user-friendly way (table?).

# Upload history item structure

Suggested information to save for an upload:

- UNIX timestamp
- Resulting reference
- Which postage batch was used
- Uploaded file or folder path (nullable, `null` if uploading from stdin)
- Type: was it a file, folder, or data from stdin? (maybe can be merged with previous field)
- Resulting feed address (nullable, only for feed upload)
- Which identity was used (nullable, only for feed upload)
- `process.argv` - Not sure about this, does it add value? If we go ahead with this, only print with `history list --verbose`

# Possible improvements

- Filter options (CLI flags) for `history list`, e.g. filter by stamp, by date, or find by name match.

# Next iteration

- Users who do not wish to have any information stored on their disk about their interactions with the network may find this feature hostile. Before releasing this, we may need to iterate on this to make this feature opt-in.
- Download history too?

slapec93:
Introducing history tracking for uploads, it gets saved in JSON format in a file called `upload-history.json` in the `.swarm-cli` folder. Also to show the history a new command group is added with two commands, `list` and `show`:

```bash
> swarm-cli history list
```
image

```bash
> swarm-cli history show 3
```
image


src/command/history/list.ts
```diff
@@ -0,0 +1,34 @@
+import { HistoryCommand } from './history-command'
+import { getHistory } from '../../service/history'
+import Table from 'cli-table3'
+import { LeafCommand } from 'furious-commander'
+import { ellipsis } from '../../utils/text'
+
+export class List extends HistoryCommand implements LeafCommand {
+  public readonly name = 'list'
```
Cafe137, line 9: Please add alias `ls`

src/command/history/show.ts
```diff
@@ -0,0 +1,46 @@
...
+export class Show extends HistoryCommand implements LeafCommand {
+  public readonly name = 'show'
+
+  public readonly description = 'Get upload history item'
+
+  @Argument({
+    key: 'index',
+    description: 'Index of the history item',
+    required: true,
+    autocompletePath: true,
```
Cafe137: Remove, this is not a path.

src/command/history/show.ts
```diff
@@ -0,0 +1,46 @@
...
+  })
+  public index!: string
```
Cafe137: You can add `type: 'number'` in `@Argument`, and type it as `public index!: number` instead.

You can remove the `parseInt` on L23 too.

Cafe137 (changes requested):

src/service/history/types/history-item.ts
Cafe137, line 1: File name should be `history-item.ts`, e.g. `withdraw-all.ts` and `cheque-command.ts`; source files in this repository follow this convention.

src/command/upload.ts
```diff
@@ -180,6 +182,18 @@ export class Upload extends RootCommand implements LeafCommand {
...
 
+    const history = new History(this.commandConfig.configFolderPath, this.console)
+
+    if (!usedFromOtherCommand) {
+      history.addItem({
+        timestamp: Date.now(),
+        reference: swarmHash,
+        stamp: this.stamp,
+        path: this.path,
+        uploadType: this.path ? 'file' : 'stdin',
```
Cafe137: `this.path` does not necessarily mean it is a `file` upload, it can still be a folder.

Please see and use `uploadingFolder` variable in `src/command/upload.ts`

(`const uploadingFolder = !this.stdin && FS.statSync(this.path).isDirectory()`)

Cafe137 (approved):

PR #683, opened 2026-04-22
feat: add upload history enable, disable and status commands

### LINKED ISSUE #681: Implement opt-out action for the upload history feature
## Problem statement

Users who do not wish to have any information stored on their disk about their uploads to the network may find the upload history feature hostile. There should be a way to disable persisting this data.

## New feature

- `swarm-cli history disable` command that disables the feature and asks the end user whether they want to delete their existing history.
- `swarm-cli history enable` command that enables this feature when disabled.
- `swarm-cli history status` command that displays whether this feature is enabled or not, and maybe prints the length and path to the history file (so if I disable it, I can double check that it is indeed removed / empty).

This setting can be saved either in `config.json`, `upload-history.json` or a new file.

## Other considerations

While it's great that the upload history feature can be disabled, the problem is that users may not know about it in the first place that their uploads are "tracked". There should be some way to inform users that their upload has been saved to history. Maybe `history.addItem` can check if it is the first save, and only then print an informative message, for example:

> The resulting Swarm reference has been saved to your upload history. You can view past uploads with `swarm-cli history list`. You can disable this feature by running `swarm-cli history disable`.

slapec93:
Add new commands to `history` group to be able to opt-out of history tracking. `enable`, `disable` and `status` got added. A new config file field (`historyEnabled`) is stored which indicates, whether we should save history items or not.


src/command/history/status.ts
```diff
@@ -0,0 +1,30 @@
...
+    this.console.log(`Status: ${statusText}`)
+
+    if (this.commandConfig.config.historyEnabled) {
+      this.console.log(`History file path: ${this.commandConfig.getHistoryFilePath()}`)
+      let size = 0
+
+      if (existsSync(this.commandConfig.getHistoryFilePath())) {
+        size = statSync(this.commandConfig.getHistoryFilePath()).size
+      }
+      this.console.log(`History file size: ${size} bytes`)
```
Cafe137: My bad with the specs - this should display the number of items/records/entries in the upload history, not size in bytes 🙂

src/command/root-command/command-config.ts
```diff
@@ -52,6 +54,24 @@ export class CommandConfig {
...
+    this.config.historyEnabled = true
+    this.saveConfig()
+  }
+
+  public disableHistory(): void {
+    this.config.historyEnabled = false
+
+    if (existsSync(this.getHistoryFilePath())) {
+      unlinkSync(this.getHistoryFilePath())
+    }
```
Cafe137: This shouldn't always delete the history.

As per of the original issue:

> command that disables the feature and asks the end user whether they want to delete their existing history

Use case 1 -- I plan to upload some test files, and for that I pause the upload tracking with this command. The command asks me if I want to delete my history, I type `n` as I don't want to lose my history. I re-enable the tracking after I am done with the test files.

Use case 2 -- I upload confidential data to Swarm and I only realize halfway through the operation that my upload history is enabled and recording these actions. I run `history disable` and type `y` to erase my history as I don't want to store data on my computer that links my actions.

Cafe137 (changes requested):
Code LGTM, only 2 comments left regarding the behavior of the commands.

src/command/history/enable.ts
```diff
@@ -0,0 +1,23 @@
...
+  public run() {
+    super.init()
+
+    if (this.commandConfig.config.historyEnabled) {
+      this.console.log(warningText('Upload history tracking is already enabled'))
+
+      return
+    }
+    this.commandConfig.setHistoryEnabled(true)
+    writeFileSync(this.commandConfig.getHistoryFilePath(), JSON.stringify([]))
```
Cafe137: Looks like it can overwrite an existing history with an empty history when resuming after a pause

Cafe137 (changes requested):
Please check the aforementioned overwrite problem, otherwise LGTM

Cafe137 (approved):

PR #685, opened 2026-04-23
feat: chunked upload

agazso:
Basic chunked upload. Not compatible with `--act`, `--encrypt`, `--redundancy` and `--sync` flags.


Cafe137:
https://github.com/ethersphere/swarm-cli/issues/778 converted this to an issue so we can cover and integrate it better in the future.

PR #686, opened 2026-04-24
feat!: add unit flag for cheque deposit and withraw

### LINKED ISSUE #684: It is hard to specify BZZ amount in `cheque deposit` and `cheque withdraw` commands
## Context

Bee nodes have a **wallet** and a **chequebook**. Funds on the **wallet** can be manually spent on postage batches and staking, while funds deposited to the **chequebook** are automatically spent by the Bee node to pay its peers for their bandwidth.

The user may deposit and withdraw funds from their wallet to the chequebook (or vice versa) at any time. To support this functionality, `swarm-cli` has the `cheque` command group with commands `deposit` and `withdraw`.

## Problem statement

The `deposit` and `withdraw` command are hard to use, because they expect the amount to be specified in PLUR. For example, naively running `swarm-cli cheque deposit 1` will deposit `0.0000000000000001` xBZZ to the chequebook which is completely useless. A recommended initial deposit to the chequebook is 1 xBZZ. To achieve that, the end user would have to type `swarm-cli cheque deposit 10000000000000000` which is obviously unreasonable.

## Enhancement

Come up with a UX improvement that helps end users specify the value they intend to use. For example, this has been addressed in the `swarm-cli stake deposit` command, where there are two mutually exclusive flags: 

```ts
@Option({
    key: 'bzz',
    description: "Amount of BZZ to add to the node's stake",
    type: 'string',
    conflicts: 'plur',
    required: true,
})
public amountBzz!: string | undefined

@Option({
    key: 'plur',
    description: "Amount of PLUR to add to the node's stake",
    type: 'bigint',
    minimum: BigInt(1),
    conflicts: 'bzz',
    required: true,
})
public amountPlur!: bigint | undefined
```

Of course it is an option to use the same approach in the `deposit` and `withdraw` commands, but some prior research should be done whether this is the best solution to the problem.

## Info

`bee-js` exposes the `BZZ` class which has convenient methods for parsing decimal strings (`'0.1'`) or number strings (`'10000000000000000'`) as BZZ and work with them more easily:

```ts
if (this.amountPlur) {
  await this.deposit(BZZ.fromPLUR(this.amountPlur))
} else if (this.amountBzz) {
  await this.deposit(BZZ.fromDecimalString(this.amountBzz))
}
```

slapec93:
Add a new flag called `--unit` for cheque deposit and withdraw command, which the user can use to pick between bzz and plur, when specifying amount


src/command/cheque/deposit.ts
```diff
@@ -11,18 +13,37 @@ export class Deposit extends ChequeCommand implements LeafCommand {
 
   @Argument({
     key: 'amount',
+    description: 'Amount of tokens to deposit',
     type: 'bigint',
-    description: 'Amount of tokens to deposit in PLUR',
     required: true,
     minimum: BigInt(1),
   })
   public amount!: bigint
```
Cafe137: This probably needs to have type `string`

With type `bigint` it probably won't accept decimal values such as `0.5` (truncates to 0, invalid) or `1.5` (loses decimal part)

src/command/cheque/withdraw.ts
```diff
@@ -12,17 +14,36 @@ export class Withdraw extends ChequeCommand implements LeafCommand {
   @Argument({
     key: 'amount',
     type: 'bigint',
-    description: 'Amount of tokens to withdraw in PLUR',
+    description: 'Amount of tokens to withdraw',
     required: true,
     minimum: BigInt(1),
```
Cafe137: This probably needs to have type `string`

With type `bigint` it probably won't accept decimal values such as `0.5` (truncates to 0, invalid) or `1.5` (loses decimal part)

Cafe137 (dismissed):

Cafe137:
Please check manually with decimal values. For example try depositing 0.5 BZZ to the chequebook and then withdrawing it back to the node wallet.

src/command/cheque/withdraw.ts
```diff
@@ -11,17 +13,49 @@ export class Withdraw extends ChequeCommand implements LeafCommand {
...
+  })
+  public unit!: string
 
   public async run(): Promise<void> {
     super.init()
 
-    const response = await this.bee.withdrawTokens(this.amount.toString())
+    const amountBzz = this.unit === 'bzz' ? BZZ.fromDecimalString(this.amount) : BZZ.fromPLUR(this.amount)
+
+    process.stderr.write(JSON.stringify(amountBzz) + '\n')
```
Cafe137: Accidental?

test/misc/monetary-units.spec.ts
```diff
@@ -38,12 +38,12 @@ describeCommand('Test Monetary units', ({ consoleMessages }) => {
...
     await invokeTestCli(['cheque', 'withdraw', '--help'])
-    expectSubstringsPrinted('amount', 'in PLUR')
+    expectSubstringsPrinted('amount', 'Amount of tokens to withdraw')
   })
 
   it('should show units in help: cheque deposit', async () => {
     await invokeTestCli(['cheque', 'deposit', '--help'])
-    expectSubstringsPrinted('amount', 'in PLUR')
+    expectSubstringsPrinted('amount', 'Amount of tokens to deposit')
   })
```
Cafe137: These two tests can now be removed

Cafe137 (changes requested):
Almost there 😌

Cafe137 (approved):
There was a test failure, if that passes we can merge

PR #696, opened 2026-04-28
chore: update default rpc url for quickstart config

### LINKED ISSUE #690: Update `quickstart` configuration file with working `resolver-options`
Suggested new value: `https://ethereum-rpc.publicnode.com`

slapec93:


Cafe137 (changes requested):
This value should be set for `resolver-options` (which is `https://cloudflare-eth.com` previously)

The Bee node has one required and one optional JSON-RPC:

- `blockchain-rpc-endpoint` - This one is for the Gnosis connection, required for all Swarm related functionality (postage batches, chequebook, staking, etc.)
- `resolver-options` - This one is for Ethereum connection to be able to resolve ENS (.eth) names. ENS lives on Ethereum mainnet, so even though everything else is on Gnosis, we need an Ethereum RPC too.

So `xdai.fairdatasociety.org` is still correct for `blockchain-rpc-endpoint`; we need to update `resolver-options` instead

Cafe137 (approved):

PR #707, opened 2026-05-04
feat: use amount argument and unit option for stake deposit

### LINKED ISSUE #704: Update the `stake deposit` command to be consistent with new `cheque` behavior
The goal is to have the same unit handling as with `cheque withdraw`, `cheque deposit`

Swarm-CLI should have a unified UX across all commands

slapec93:
Use the same pattern for the stake deposit command, which was recently introduced for `cheque deposit` and `cheque withdraw`


src/command/stake/deposit.ts
```diff
@@ -1,9 +1,10 @@
 import { BZZ } from '@ethersphere/bee-js'
-import { LeafCommand, Option } from 'furious-commander'
+import { Argument, LeafCommand, Option } from 'furious-commander'
 import { exit } from 'process'
 import { createSpinner } from '../../utils/spinner'
 import { RootCommand } from '../root-command'
 import { VerbosityLevel } from '../root-command/command-log'
+import { Context } from 'vm'
```
Cafe137: Wrong import

src/command/stake/deposit.ts
```diff
@@ -12,33 +13,49 @@ export class Deposit extends RootCommand implements LeafCommand {
...
+          if (amount <= BigInt(0)) {
+            return [`Invalid amount '${value}'. Amount must be a positive integer.`]
+          }
+        } catch (e) {
+          return [`Invalid amount '${value}'. Amount must be a positive integer.`]
+        }
+      }
+
+      return []
+    },
```
Cafe137: Repeated 3 times if I am not mistaken.
Maybe extract this as a stateless utility function to a new `src/utils/validate.ts`?

Cafe137 (approved):

PR #708, opened 2026-05-05
feat: show warning when new swarm-cli version is available

### LINKED ISSUE #345: Warning about new version
It is I think quite expected that swarm-cli will be installed into `--global` scope and it might be hard to keep track of what version of swarm-cli I am running and if there is not a new one out. I would maybe suggest having a check on every run that would check if a new version is not out and if so then display a warning about the new version (of course not in quite mode).

slapec93:
image


Cafe137:
@slapec93 what happens if the version check network request fails? My concern is that it prints an ugly error - can we handle it with a "Failed to check swarm-cli version"?

Cafe137 (dismissed):
Code LGTM but please see comment

Cafe137 (approved):

PR #715, opened 2026-05-13
test: decrease coverage

slapec93:


slapec93:
The new approach works 🎉

PR #719, opened 2026-05-18
feat: add qr option for upload command

### LINKED ISSUE #269: Feature idea: show QR code in terminal
The result of many operations is a very long hash or url, which can be cumbersome if you want to transfer it to an other device (e.g. mobile). It would be good to have an option to print QR codes as well.

slapec93:
Add a new flag `--qr`, what the user can use to have a QR code as the output an upload or feed upload

image


src/command/feed/feed-command.ts
```diff
@@ -1,6 +1,8 @@
 import { Wallet } from '@ethereumjs/wallet'
 import { Reference, Topic } from '@ethersphere/bee-js'
 import { Option } from 'furious-commander'
+import chalk from 'chalk'
+import QRCode from 'qrcode'
```
Cafe137: Remove unused imports

src/command/upload.ts
```diff
@@ -16,6 +17,8 @@ import { createKeyValue, warningSymbol, warningText } from '../utils/text'
 import { RootCommand } from './root-command'
 import { VerbosityLevel } from './root-command/command-log'
 import { History } from '../service/history'
+import chalk from 'chalk'
```
Cafe137: Remove unused imports

Cafe137 (changes requested):
Thanks, functionality is perfect, but please do not introduce new check warnings

Cafe137 (approved):

PR #752, opened 2026-06-04
fix: generate qr from a publicly available url

### LINKED ISSUE #751: Use gateway URL for generating QR code
The current implementation uses localhost for the QR code, which is not useful, change it to use the gateway url

slapec93:


Cafe137 (changes requested):
Please add a test that confirms that the QR function was called with a `api.gateway.ethswarm.org/bzz/[0-9a-f]{64}` given that the upload was local

Cafe137 (approved):

PR #775, opened 2026-07-10
chore: strip down eslint and fix violations

### LINKED ISSUE #779: `manifest` commands fail because of erasure coding
The `manifest sync` command hashes contents of a local folder locally, and compares the hash against a remote mantaray manifest to only update the delta.

This started failing since the Bee release where uploads are now erasure coded by default. We need a `swarm-core` so that we can hash locally with erasure code support.

slapec93:


test/command/pss.spec.ts
```diff
@@ -68,6 +69,7 @@ describeCommand('Test PSS command', ({ getNthLastMessage, getLastMessage }) => {
       'test/testconfig/in.txt',
       ...getStampOption(),
     ])
+    await receivePromise
     await System.sleepMillis(4000)
```
Cafe137: Off: Do we need to keep the original 4s sleep if we now await receive?

slapec93: probably not 🤔

test/custom-matcher.ts
```diff
@@ -23,7 +23,7 @@ export function toMatchLinesInOrder(received: string[], pattern: string[][]) {
 
 export function toMatchLinesInAnyOrder(received: string[], pattern: string[][]) {
   for (const p of pattern) {
-    if (received.some(r => r.includes(p[0]))) {
+    if (received.some(r => r.includes(p[0] ?? ''))) {
```
Cafe137: Wait... Why are we checking only `p[0]`? This whole matcher seems wrong

Cafe137 (approved):

PR #777, opened 2026-07-14
fix: disable ANSI colors in history list when stdout is not a TTY

### LINKED ISSUE #771: Piping `history list` to `less` is hard to read -- disable ANSI colors when `process.stdout.isTTY` is not `true`
I have a long upload history, so I piped it to `less` to navigate it better. However, the output was hard to read due to the ANSI escape codes. 

Most CLI apps only enable colors when `stdout` is a TTY, and disable them when `stdout` is not a TTY.

Please apply this fix to the `history list` command so that it doesn't use colors when piped to another command.

Image

Riki0923:


src/command/history/list.ts
```diff
@@ -22,10 +22,14 @@ export class List extends HistoryCommand implements LeafCommand {
...
     }
+
+    const useColors = Boolean(process.stdout.isTTY)
+
     const table = new Table({
       head: ['Index', 'Timestamp', 'Reference', 'Postage stamp batch ID', 'File path', 'Upload type'],
       style: {
-        head: ['green', 'bold'],
+        head: useColors ? ['green', 'bold'] : [],
+        border: useColors ? ['grey'] : [],
```
slapec93, line 32: can you please cover the `isTTY = false` case with spec(s)?

slapec93 (changes requested):
Provide some tests for the new functionality please, otherwise LGTM 👌

src/command/history/list.ts
Riki0923: Added tests to history.spec.ts file

test/command/history.spec.ts
```diff
@@ -2,6 +2,9 @@ import chalk from 'chalk'
 import { randomUUID } from 'crypto'
 import { describeCommand, invokeTestCli } from '../utility'
 import { getStampOption } from '../utility/stamp'
+import colors from '@colors/colors/safe'
+
+const ANSI_PATTERN = /\u001B\[\d+m/ //adding this for testing the ansi disable part
```
slapec93: total nit picking, but since this one is only related to the test cases L131-L166, I would move this const to that block

slapec93 (commented):
Just one more suggested improvement, I'm okay without it

test/command/history.spec.ts
Riki0923: Done

Cafe137 (approved):

slapec93 (approved):
👌 ✅

PR #791, opened 2026-09-10
ci: accept bee-released repository_dispatch

darkobas2:
`update-bee-version.yaml` can only be started by hand right now, so the Bee version in `quickstart.ts` has never been bumped on its own.

This adds a `repository_dispatch` trigger for the `bee-released` event that ethersphere/bee sends when a stable Bee tag is pushed, reading the version from `client_payload`. The existing `Validate version` step still rejects anything malformed.

Merge this before ethersphere/bee#NNN, which sends the event.


darkobas2:
Sender side: ethersphere/bee#5607. Merge this one first.

PR #792, opened 2026-09-14
feat: prompt user for stamp type, when flag not set

### LINKED ISSUE #790: `stamp create` command should prompt for type (immutable or mutable) rather than defaulting to one

slapec93:
image


Cafe137 (approved):
Perfect, thanks

